PR #41, opened 2016-03-11
Solves UnicodeEncodeError for 'charmap' codec for unicode output

suspended:


suspended:
Not sure why the CI build failed. It should work though

jsstylos:
Thanks for the pull request.

I think the integration test is failing because of environment issues, not related to the change.

The `.encode('utf-8')` prints as expected on Python 2 but on Python 3 encodes the string as bytes: `b'...\n \n \n'`. I did a quick check but haven't yet found the recommended way of printing utf strings in a Python 2 and 3 compatible way. Does anyone know the best way to do this while supporting Python 2 and 3?

jsstylos:
Looks like `.content.decode('utf-8')` works as expected in both Python 2 and 3. @suspended Does this work for your use case?

suspended:
`.decode('utf-8')` works fine for me. I am using Python 2.

jsstylos:
Ok, great. I was going to check this in but our integration tests are
failing because of Natural Language Classifier uptime issues, so I've been
holding up releasing until that's fixed.

jsstylos:
Fixed in f0d5865c6f28289f671809019ad21b0ee5e722f7

Issue #40, opened 2016-03-09
Instantiating ToneAnalyzerV3Beta should have default version value


peterrickwood:
The example code suggests you can instantiate a tone analyzer by passing in just 2 arguments, as follows: 

`ta = ToneAnalyzerV3Beta(username='yourusername',password='yourpassword')`

However, this is not right -- you have to pass in a version argument as well, thusly:

`tone_analyzer = ToneAnalyzerV3Beta(version='2016-02-11', username='yourusername', password='yourpassword')`

Since it doesnt seem obvious what the right version argument should be, maybe this should default to ToneAnalyzerV3Beta.latest_version?

jsstylos:
This is a tricky design question that we've looked at before but may have to revisit.

The challenge is that version dates are for describing backwards-incompatible changes, and setting a default version date in the SDK makes every version date change a breaking change for the SDK, so that update the SDK could break your code.

This is similar to the motivation for not having default version dates at the service API level: https://github.com/watson-developer-cloud/api-guidelines#why-make-the-version-date-required-wouldnt-it-be-simpler-to-make-it-optional. The SDK is potentially easier to address, but is still challenging, especially since it combines so many different separate services that are each independently versioned. The combination of multiple, separately versioned services is one difference of this SDK to other SDKs that wrap a single version-dated service.

For now the recommended (and enforced) use of the SDK is to provide a version date when constructing a service object. We make the sample code and document use the latest version, so users should copy and paste the code at authoring and update is as desired. In exchange, they shouldn't have to worry that upgrading the SDK might break their code, even when we introduce new version-dates.

PR #42, opened 2016-03-14
Fixed URL issue for document conversion

EpicureanConslt:
Simple, straightforward change. The document conversion API URL is incorrect and does not work. I have added the version parameter to the URL, to fix this issue.

Old URL Format: /v1/convert_document
New URL Format: /v1/convert_document?version=2015-12-15


jsstylos:
Hi, thanks for the pull request. The version is a required parameter of the (non-experimental) service (see: https://github.com/watson-developer-cloud/python-sdk/issues/40#issuecomment-194531239). The service should used the specified version date and not a hard-coded one in the SDK.

The experimental version of the service is no longer available, and will be removed from a future version of the SDK.

PR #52, opened 2016-03-30
Adding example script for running learning-to-rank experiments

rchaks:
Demonstrate use of the /rank API methods (part of the Retrieve and Rank service) using a command line runnable script for learning-to-rank experiments.  Blog containing detailed experimental usage of the script to follow once code becomes available in the public repo.

### DIFF
--- a/examples/rank_with_evaluation_v1.py
+++ b/examples/rank_with_evaluation_v1.py
@@ -1,0 +1,499 @@
+"""
+Script for generating RaaS predictions.  Refer to the following wiki for details:
+ https://w3-connections.ibm.com/wikis/home?lang=en-us#!/wiki/Wc649e283e9c4_46b5_8fc1_7b91b8004716/page/How%20to%20run%20RaaS%20experiments%20using%20Bluemix%20%28requires%20acces%20to%20%60Retrieve%20and%20Rank%60%20service%20in%20bluemix%29
+"""
+import argparse
+import csv
+import json
+import logging
+import math
+import sys
+import tempfile
+from collections import defaultdict
+from time import sleep
+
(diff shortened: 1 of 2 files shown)


rchaks:
@germanattanasio - Do we need to check in unit tests for example scripts (there didn't seem to be any for previous example scripts)?  If so, are there prior examples scripts I could take a look at?  Specifically, I would like to figure out how to mock out the REST calls (my local unit tests required a valid set of credentials that could be used to call the RnR service).

germanattanasio:
Python is pretty flexible, if you look at our tests we use `responses` so you can mock the API and call your script using `os.processor` like if it was a script call

jsstylos:
I added the file in a folder called `utilities` in 302c9b588610536ec99f6c6673799c3d68789cdb

rchaks:
@jsstylos - thanks for pulling that in!  One note: I saw that you re-named it a **retrieve-and-**rank utility.  However, the utility only really allows the user to use the **rank** api methods.  The goal being to show the use of the learning-to-rank as a standalone algorithm for solving non-retrieval tasks (blog upcoming with real example that uses the script).

jsstylos:
@rchaks: The naming is intended to show the service that is used, not imply that it's using all of the capabilities of the service; just calling it rank didn't make it clear which Developer Cloud service it is using. I imagine that a speech utility would be speech_to_text_v1_websockets.py or something, even if it wasn't using all of the capabilities of the Speech to Text service, and we might have separate speech_to_text_v1_customization.py and retrieve_and_rank_v1_retrieve.py utilities. An alternative design would be to create a folder to act as namespace, /utilities/retrieve_and_rank_v1/rank.py

Issue #53, opened 2016-04-07
AlchemyNewsDataV1 does not support pagination


jtreleaven:
Will currently throw the following error if you try to add the "next" query parameter to retrieve the next page. 
```
Traceback (most recent call last):
  File "/Library/Frameworks/Python.framework/Versions/2.7/lib/python2.7/site-packages/watson_developer_cloud/alchemy_data_news_v1.py", line 39, in get_news_documents
    return self._alchemy_html_request(method_url='/data/GetNews', method='GET', params=params)
  File "/Library/Frameworks/Python.framework/Versions/2.7/lib/python2.7/site-packages/watson_developer_cloud/watson_developer_cloud_service.py", line 177, in _alchemy_html_request
    accept_json=True)
  File "/Library/Frameworks/Python.framework/Versions/2.7/lib/python2.7/site-packages/watson_developer_cloud/watson_developer_cloud_service.py", line 245, in request
    raise WatsonException('Error: ' + error_message)
WatsonException: Error: Invalid field = 'next'
```
I've identified the bug in the code and will place a PR soon. Adding this issue for documentation/tracking purposes.

jsstylos:
Fixed in 302c9b588610536ec99f6c6673799c3d68789cdb

PR #54, opened 2016-04-07
added support for pagination for the Watson News Data api.

jtreleaven:
Added a next_page parameter to allow for pagination to work with minimal changes to the get_news_documents function. This PR is in response to the issue I created about the lack pagination support #53.

### DIFF
--- a/watson_developer_cloud/alchemy_data_news_v1.py
+++ b/watson_developer_cloud/alchemy_data_news_v1.py
@@ -36,5 +36,8 @@ class AlchemyDataNewsV1(WatsonDeveloperCloudService):
                   'next': next_page}
         if isinstance(query_fields, dict):
             for key in query_fields:
-                params[key if key.startswith('q.') else 'q.' + key] = query_fields[key]
+                if key.startswith('enriched.url') or key.startswith('original.url'):
+                    params[key if key.startswith('q.') else 'q.' + key] = query_fields[key]
+                else:
+                    params[key] = query_fields[key]
         return self._alchemy_html_request(method_url='/data/GetNews', method='GET', params=params)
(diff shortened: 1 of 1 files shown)


jsstylos:
Good catch!

Can you add the field like the other params (without checking to see if it's a string type, since it already gets run through a "remove none" step):

```        params = {'start': start,
                  'end': end,
                  'maxResults': max_results,
                  'return': return_fields,
                  'timeSlice': time_slice,
                  'next': next_page}```

jtreleaven:
Sure thing! I forgot the html_request_method striped out none values!

jsstylos:
What parameter values for the `query_fields` array are the latest changes made to support? From the API reference at https://alchemyapi.readme.io/v1.0/docs/rest-api-documentation it wasn't clear to me that all of the q. prefixed parameters would be limited to the q.enriched.url and q.original.url, or that there were additional non-q.-prefixed parameters other than outputMode, start, end, timeSlice and maxResults.

The `key.startswith('enriched.url') or ...` makes the `if key.startswith('q.')` check unnecessary, if we do want to switch to code to only prefix q. for the enriched.url.* and original.url.* parameters.

jsstylos:
Ok, I found https://github.com/watson-developer-cloud/python-sdk/issues/55 with examples of `dedup` and `dedupThreshold`. I'm still not sure if the right solution is to add optional solution is to add optional fields for `dedup` and `dedupThreshold` and leave `query_fields` specifically for the `q.` query parameters or not.

jtreleaven:
I agree! I wasn't quite sure what to do about the growing number of non 'q.' prefixed fields.
What do you think about creating an option_fields parameter that can contain non 'q.' prefixed url params?

jsstylos:
I'm not sure about what the best way to support additional non-q. fields is. On the one hand they're not unbounded (like the q. fields are), but still numerous, and hard to track down since they're not all documented in one place.

Having a catch-all "option_fields" would be one solution (and is similar to the approach taken in the Java SDK), although it makes discovery more difficult for any fields that aren't also called out as explicit stand-alone parameters. Using `**kwargs` would be a Python specific way, although it has some of the same issues.

For now I added explicit parameters for `dedup`, `dedupThreshold` and `rank`, leaving the `query_fields` specifically for the `q.` parameters`.

jsstylos:
I added the code in 302c9b588610536ec99f6c6673799c3d68789cdb

jtreleaven:
Looks good to me! Thanks for the quick responses and help!

PR #64, opened 2016-06-09
Added timestamp parameter to SpeechToTextV1.recognize

madra:
I needed an easy way to get back time-stamps from  SpeechToTextV1.recognize()
I added recognize(self, audio,
                  content_type, continuous=False,
                  model=None, timestamps=False)
timestamps=True adds {'timestamps':True} to the headers


kognate:
Can you provide some details why this timestamp param is important to return from this method?  Also, why not just subclass `SpeechToTextV1` yourself rather than modify this class?

germanattanasio:
I've added the missing parameters including timestamp in #67

jsstylos:
Thanks for the report! This is fixed in the latest release.

PR #70, opened 2016-06-22
Updates Tone Analyzer to GA version: latest api(2016-05-19) and 'params' arg

herchu:
This updates the latest minor API version, and allows passing a 'params' argument (to specify for example sentences=false or tones=social args).

It also adds 2 testcases for Tone; there was none.

### DIFF
--- a/watson_developer_cloud/tone_analyzer_v3.py
+++ b/watson_developer_cloud/tone_analyzer_v3.py
@@ -19,7 +19,6 @@ The v3 Tone Analyzer service
 
 
 from watson_developer_cloud.watson_developer_cloud_service import WatsonDeveloperCloudService
-import copy
 
 class ToneAnalyzerV3(WatsonDeveloperCloudService):
     default_url = 'https://gateway.watsonplatform.net/tone-analyzer/api'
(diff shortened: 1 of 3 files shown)


watson_developer_cloud/tone_analyzer_v3.py
```diff
@@ -19,24 +19,25 @@
...
-
+import copy
 
 class ToneAnalyzerV3(WatsonDeveloperCloudService):
     default_url = 'https://gateway.watsonplatform.net/tone-analyzer/api'
-    latest_version = '2016-02-11'
+    latest_version = '2016-05-19'
 
-    def __init__(self, version, url=default_url, username=None, password=None, use_vcap_services=True):
+    def __init__(self, version=latest_version, url=default_url, username=None, password=None, use_vcap_services=True):
```
germanattanasio: @herchu we don't use a default `version_date`. It's a required parameter and users should always provide a valid value

watson_developer_cloud/tone_analyzer_v3.py
```diff
@@ -19,24 +19,25 @@
...
-    def tone(self, text):
+    def tone(self, text, params=None):
         """
         The tone API is the main API call: it analyzes the "tone" of a piece of text. The message is analyzed from
         several tones (social tone, emotional tone, writing tone), and for each of them various traits are derived
         (such as conscientiousness, agreeableness, openness).
         :param text: Text to analyze
         """
-        params = {'version': self.version}
+        args = copy.copy(params) if params is not None else {}
```
germanattanasio: this looks complicated. The SDK already adds the version to the query. Look at Document conversion https://github.com/watson-developer-cloud/python-sdk/blob/master/watson_developer_cloud/document_conversion_v1.py

herchu:
@germanattanasio I have done some fixes. Can you review?

PR #76, opened 2016-07-05
[document-conversion] Adds index document API

samir-patel:
This commit adds the index_document API for the document
conversion service for the python SDK. It also includes
updated examples for using the index_document API.


jsstylos:
I integrated this in the latest release -- thanks for the code!

PR #85, opened 2016-07-27
R&R and conversation unit tests

Mikemosca:
Add a R&R unit test
Add conversation unit test


test/test_retrieve_and_rank_v1.py
```diff
@@ -0,0 +1,225 @@
...
+
+    listrank_url = 'https://gateway.watsonplatform.net/retrieve-and-rank/api/v1/rankers'
+    listrank_response = '{"rankers":[{"ranker_id":"3b140ax14-rank-10383","url":"https://gateway.watsonplatform.net/retrieve-and-rank' \
+                    '/api/v1/rankers/3b140ax14-rank-10383","name":"pythonRank","created":"2016-07-15T19:11:27.801Z"}]}'
+                    
+    responses.add(responses.GET, listrank_url,
+              body=listrank_response, status=200,
+              content_type='application/json')
+
+    retrieve_and_rank.list_rankers()
```
germanattanasio: if you save the `list_rankers` result into a variable you can then check that the number of rankers match the static response.

```python
ranker_list = retrieve_and_rank.list_rankers()
assert len(ranker_list['rankers']) == 1
```

germanattanasio:
@Mikemosca : The unit test seems OK but the integration tests need some work so I would suggest you to remove the integration test for now so that we can merge this pull request and we can work on the integration tests in the next sprint. I think we need to add support for the integration tests but we should not have a new folder called `testintegration`. We need to add those tests into the `test` folder and use `pytest` annotation so mark them as integration test. I can help you with that next week

Mikemosca:
@germanattanasio: thanks so much.  made the changes (removed integration test for now and added ranker_list check).  Now I am confused about how to get these changes into the pull request.

PR #91, opened 2016-07-29
[R&R] Remove type checking in cluster creation fix #87

### LINKED ISSUE #87: When creating a solr cluster passing in a cluster size fails.
When creating a solr cluster passing in a cluster size fails.  It may be a case where we just need to import long
```
retrieve_and_rank.create_solr_cluster(cluster_name='pythonCluster', cluster_size=10)
```
```
self = <watson_developer_cloud.retrieve_and_rank_v1.RetrieveAndRankV1 object at 0x0000000003BC1940>
cluster_name = 'pythonCluster', cluster_size = 10

    def create_solr_cluster(self, cluster_name=None, cluster_size=None):
>       if cluster_size and isinstance(cluster_size, (int, long)):
E       NameError: name 'long' is not defined

File "C:\Users\IBM_ADMIN\Documents\Work\Watson\DevExp\python-sdk\watson_developer_cloud\retrieve_and_rank_v1.py", line 36
NameError    
```

germanattanasio:
Since python 3 doesn't have long we can't check if `cluster_size` is a long instance.

I did a quick check and 

if `cluster_size` is not null we can just convert it to string. 

```
>>> str(3L)
'3'
>>> str(3)
'3'
>>> str('3')
```

### DIFF
--- a/watson_developer_cloud/retrieve_and_rank_v1.py
+++ b/watson_developer_cloud/retrieve_and_rank_v1.py
@@ -32,7 +32,7 @@ class RetrieveAndRankV1(WatsonDeveloperCloudService):
         return self.request(method='GET', url='/v1/solr_clusters', accept_json=True)
 
     def create_solr_cluster(self, cluster_name=None, cluster_size=None):
-        if cluster_size and isinstance(cluster_size, (int, long)):
+        if cluster_size:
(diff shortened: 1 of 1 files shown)


germanattanasio:
This fixes #87

PR #130, opened 2016-12-06
adds (some) unit tests for alchemy language

kognate:
This adds some unit tests

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,6 +2,9 @@ language: python
 sudo: false
 notifications:
   email: false
+addons:
+  apt_packages:
+    - pandoc
 python:
 - '2.7'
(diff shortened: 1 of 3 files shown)


jsstylos:
Do we need sudo and pypandoc on Travis? Right now it's only used when publishing, which Travis doesn't currently need to do.

kognate:
The builds failed without it when it did the `python setup.py egg_info` step

jsstylos:
I think that was due to a related issue that was fixed in https://github.com/watson-developer-cloud/python-sdk/pull/132 without sudo or pypandoc -- the builds are working now.

kognate:
replaced with #133

PR #137, opened 2016-12-27
add create/update/delete methods for configurations in discovery v1

kognate:
These methods were left out because tooling doesn't
support them very well and they are likely to change but
we got a request to add them for now.


kognate:
This branch is messed up.

PR #139, opened 2016-12-30
Now a document can be added from a string

### LINKED ISSUE #140: WDS add document only allows files
The add_document method for discovery only allows file objects.  It should also allow the addition of data in a string buffer as well as file objects.

kognate:
Before you could only add documents as files. now
the document data doesn't have to be in a file, it can be
in a string.

Fixes #140

### DIFF
--- a/test/test_discovery_v1.py
+++ b/test/test_discovery_v1.py
@@ -342,3 +342,18 @@ def test_document():
                               document_id='docid')
 
     assert len(responses.calls) == 5
+
+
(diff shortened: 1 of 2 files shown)


jsstylos:
Should probably clean up the parameter names to match the snake casing of the other parameters.

kognate:
I updated those param names.

jsstylos:
Looks like pylint is complaining about indentation now. metadata is a word, so doesn't need to be separated -- the others look good.

jsstylos:
Looks good

PR #143, opened 2017-01-05
Adds missing get_document

### LINKED ISSUE #142: [discovery] Add ability to check doc status
The Discovery API exposes a get doc endpoint to check the status of a document after ingestion (https://www.ibm.com/watson/developercloud/discovery/api/v1/#get-doc) but this is not available via the sdk.

kognate:
The get document method wasn't implemented. This
adds it.  Fixes #142

### DIFF
--- a/test/test_discovery_v1.py
+++ b/test/test_discovery_v1.py
@@ -340,6 +340,11 @@ def test_document():
                   status=200,
                   content_type='application/json')
 
+    responses.add(responses.GET, del_doc_url,
+                  body="{\"body\": []}",
+                  status=200,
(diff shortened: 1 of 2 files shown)


jsstylos:
Pylint's complaining.

jsstylos:
A bit broader than just this PR, but the version should be passed from self.version, and should never default to latest_version (in the call or constructor), so that the SDK can be updated to support the latest (breaking) version without breaking its users or requiring a major SDK version bump.

kognate:
Do you want me to fix in this PR or create another PR to cover that?

PR #146, opened 2017-01-10
adds ability to add custom corpora

kognate:
This is part of wdc/issues/551 for adding
customization support to sdk.


jsstylos:
Looks like Travis is failing with: `No such file or directory: './resources/speech_to_text/corpus-short-1.txt'`

kognate:
Yeah, I'm fixing that path problem.  <badjoke>It works on my machine</badjoke>

jsstylos:
I think add_corpora should be add_corpus, since it's only adding one. Otherwise looks good.

PR #147, opened 2017-01-17
adds custom word customizations

kognate:
Final PR in the fixing the #551 issue


jsstylos:
Looks like Pylint warnings

jsstylos:
Should `list_custom_word` be `get_custom_word` or something else? I don't think we usually use list for retrieving one item.

kognate:
I was naming it to track with the api nomenclature.  The use 'list a custom word' and 'list custom words' to describe each api endpoint.   Change it to `get`?

jsstylos:
Yeah, I think change to get

PR #150, opened 2017-02-03
Fix line ending in files

sdague:
It looks like git on some of the developer machines wasn't set up to
do automatic line ending handling, as such a lot of litteral ^M are
checked into the respository from files that were created entirely on
Macs. When viewing these files on a Linux machine, they look all kinds
of weird because the line ending is ^J.

This creates a .gitattributes file and sets all the files with extra
^M checked into the repository to be reset to git native ^J. When
checked out locally git will do translation to the right local endings
(^M on Mac, ^M^J on Win, ^J on Linux).


jsstylos:
Good catch, thanks for the fix!

PR #151, opened 2017-02-06
Adds support for NLU

kognate:
This adds support for NLU analyze api and sentiments.

### DIFF
--- a/examples/natural_language_understanding_v1.py
+++ b/examples/natural_language_understanding_v1.py
@@ -1,24 +1,27 @@
 
 # coding: utf-8
 
-# In[1]:
+# In[4]:
 
 import sys
 import os
 sys.path.append(os.path.join(os.getcwd(),'..'))
 import watson_developer_cloud
+import watson_developer_cloud.nlu.features.v1 as features
 
 
-# In[2]:
+# In[3]:
 
 nlu = watson_developer_cloud.NaturalLanguageUnderstandingV1(version='2016-01-23',
-                                                            username='<USERNAME>',
-                                                            password='<PASSWORD>')
+                                                            username='USERNAME',
+                                                            url='https://gateway-s.watsonplatform.net/natural-language-understanding/api',
+                                                            password='PASSWORD')
 
 
-# In[10]:
+# In[6]:
 
-nlu.analyzeText(text='this is my experimental text.  Bruce Banner is the Hulk and Bruce Wayne is BATMAN! Superman fears not Banner, but Wayne.',features={'entities': {}})
+nlu.analyzeText(text='this is my experimental text.  Bruce Banner is the Hulk and Bruce Wayne is BATMAN! Superman fears not Banner, but Wayne.',
+                features=[features.Entities(), features.Keywords()])
 
 
 # In[ ]:
--- a/examples/natural_language_understanding_v1.py
+++ b/examples/natural_language_understanding_v1.py
@@ -1,0 +1,27 @@
+
+# coding: utf-8
+
+# In[1]:
+
+import sys
+import os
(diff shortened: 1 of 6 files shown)


watson_developer_cloud/natural_language_understanding_v1.py
```diff
@@ -0,0 +1,91 @@
...
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+from watson_developer_cloud.watson_developer_cloud_service import \
+    WatsonDeveloperCloudService
+import json
+
+base_url = 'https://gateway.watsonplatform.net'
+default_url = '{0}/natural-language-understanding/api'.format(base_url)
```
jsstylos: I think the other services put this inside the class

watson_developer_cloud/natural_language_understanding_v1.py
```diff
@@ -0,0 +1,91 @@
...
+        if len(featureList) < 1:
+            raise ValueError("Must supply at least one feature")
+
+        return self.request(method='POST', url='/v1/analyze',
+                            params={"version": self.version},
+                            headers={'content-type': 'application/json'},
+                            data=body,
+                            accept_json=True)
+
+    def analyzeText(self, features, text):
```
jsstylos: In AlchemyLanguage, which exposed separate endpoints for different input formats, we combined this into single SDK methods that took optional parameters for each of the input types, e.g. (url=None, text=None, html=None). For NLU, which combines the endpoints, I think we'd also want the combined SDK method.

The other services also use snake_case for the method names, parameters (and internal variables).

The service supports additional top-level parameters in addition to features and text (I think it's: url=None, clean=True, xpath=None, fallback_to_raw=True, return_analyzed_text=False, language=None). The way we've supported these before in the Python SDK is by making the top-level JSON fields parameters in the method, and by (usually) setting the service's default value for the parameters in the SDK.

watson_developer_cloud/nlu/features/v1/__init__.py
```diff
@@ -0,0 +1,72 @@
+class Feature(object):
```
jsstylos: Not sure now this works -- does it essentially type-check the valid keys in the dict they send in the features parameter?

kognate: you have to use the staging url for it to work.  Yes, you are correct, it type checks the features  param (which is a list of feature subclasses).

watson_developer_cloud/natural_language_understanding_v1.py
kognate: ok

watson_developer_cloud/natural_language_understanding_v1.py
kognate: OK,  I'll add the combined method and snake case the methods.  

The top level params I left off because not all of them work right now so I don't have a good way to test them.

examples/natural_language_understanding_v1.py
```diff
@@ -0,0 +1,30 @@
...
+sys.path.append(os.path.join(os.getcwd(),'..'))
+import watson_developer_cloud
+import watson_developer_cloud.nlu.features.v1 as features
+
+
+# In[3]:
+
+nlu = watson_developer_cloud.NaturalLanguageUnderstandingV1(version='2016-01-23',
+                                                            username='USERNAME',
+                                                            url='https://gateway-s.watsonplatform.net/natural-language-understanding/api',
```
jsstylos: In the public example we probably shouldn't use gateway-s, even though the public gateway doesn't work yet.

PR #155, opened 2017-02-07
Support for the new Preferable Options feature

annabekkerman:


annabekkerman:
@kognate I actually did add the new parameter to the Tradeoff API example call in `examples/tradeoff_analytics_v1.py`.  Now you'll see the new section that contains preferable options in the response json. 

Given that this example simply printed the output I thought that making the new section to appear in there should be enough. Did you have a more extensive example in mind?

PR #164, opened 2017-02-23
make sure to add all modules for instalation

### LINKED ISSUE #163: Error importing library
I'm using the watson_developer_cloud Python library to experiment with Watson. I have used PIP to install the watson_developer_cloud. This appears to work. When I try this:

import watson_developer_cloud.natural_language_understanding.features.v1 as features

I get an error saying: ModuleNotFoundError: No module named 'watson_developer_cloud.natural_language_understanding'

Thanks for your help

kognate:
This fixes #163 because on release these files weren't included and I was able to replicate the reporters findings only when using a clean vm to test with.

### DIFF
--- a/MANIFEST.in
+++ b/MANIFEST.in
@@ -2,5 +2,6 @@ include README.md
 include LICENSE
 
 recursive-include examples *.py *.md
+recursive-include watson_developer_cloud *.py *.md
 global-exclude .DS_Store
 global-exclude *.pyc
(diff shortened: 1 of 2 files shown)


germanattanasio:
@kognate can you revert the `bump version` commit ?

I think the way we do releases is using `bumpversion`. So we first do all the code changes and then we do``` bumpversion patch```

germanattanasio:
See https://github.com/watson-developer-cloud/python-sdk/blob/master/RELEASES.md

kognate:
@german I reverted it.

PR #173, opened 2017-03-06
changed name to have dashes instead of underscores

### LINKED ISSUE #172: NLU service name is wrong in the sdk
The NLU service name is serperated by dashses and not underscores.

kognate:
The NLU name has dash seperators.

Fixes #172


kognate:
Ignore this.  This was an error.

PR #177, opened 2017-03-20
[Do Not Merge] Very premilinary websockets

### LINKED ISSUE #12: Switch Speech to Text to use WebSockets
HI @daniel-bolanos I was thinking we can provide support for WebSockets in the sdk by adding an existing library and handling the authorization using the authorization service. 
What do you think?

We should:
- [x] Find a websocket library
- [x] Design a listener/observer pattern to model websockets

kognate:
This uses autobahn to support websockets.  This
needs a ton of review.  It works on my machine, but
I'm not super confident about it and it doesn't
have any tests yet.

Fixes #12


watson_developer_cloud/speech_to_text_v1.py
```diff
@@ -18,10 +18,79 @@
...
 import json
-
+import base64
+
+from autobahn.twisted.websocket import WebSocketClientFactory,\
+    WebSocketClientProtocol, connectWS
+from twisted.internet import ssl, reactor
+
+
+class STTWebsocketClientProtocol(WebSocketClientProtocol):
```
germanattanasio: `SpeechToTextWebSocketClientProtocol`

watson_developer_cloud/speech_to_text_v1.py
```diff
@@ -18,10 +18,79 @@
...
+            json_payload = json.loads(payload)
+            try:
+                self.messageCallback(json_payload)
+            except AttributeError:
+                # if we don't have a message callback, don't worry about it
+                pass
+            if 'results' in json_payload:
+                if len(json_payload['results']) > 0:
+                    if 'final' in json_payload['results'][0]:
+                        if json_payload['results'][0]['final']:
```
germanattanasio: I don't think we want to close the `WebSocket` when`final=true`. The user can send more audio and final just means that the server won't go back to the sentence and change it.

watson_developer_cloud/speech_to_text_v1.py
```diff
@@ -18,10 +18,79 @@
...
+                            # if we're final, close it up
+                            self.sendClose(1000)
+
+    def onClose(self, wasClean, code, reason):
+        reactor.stop()
+
+class STTInterfaceFactory(WebSocketClientFactory):
```
germanattanasio: `content_callback` needs to handle different messages:
 * onClose()
 * onConnected()
 * onListening()
 * onError()
 * onMessage() <- this is already implemented

watson_developer_cloud/speech_to_text_v1.py
```diff
@@ -61,6 +130,64 @@ def recognize(self, audio, content_type, continuous=False, model=None,
                             data=audio, params=params,
                             stream=True, accept_json=True)
 
+    def recognize_ws(self, audio, content_type,
```
germanattanasio: We should return an object with a `close()` method so that the user can close the WS

germanattanasio:
Here is an example of streaming https://github.com/crossbario/autobahn-python/tree/master/examples/twisted/websocket/streaming#streaming-processing

watson_developer_cloud/speech_to_text_v1.py
kognate: OK

watson_developer_cloud/speech_to_text_v1.py
kognate: Not closing it puts the code into a strange situation because of the twisted reactor.

watson_developer_cloud/speech_to_text_v1.py
kognate: ok

watson_developer_cloud/speech_to_text_v1.py
kognate: I'll have to look into this.  I think this will cause the reactor to leak out of the class.  Currently, it's entirely encapsulated.

jsstylos: In the Node SDK this is called `create_recognize_stream` (well, in camelCase).

kognate: ok, will change that.

kognate:
@germanattanasio yeah, I looked at those streaming examples.  The producer found in the twisted FileSender conforms to that api but I couldn't get it to do what I wanted.

watson_developer_cloud/speech_to_text_v1.py
germanattanasio: You should close it when there is no more data to be sent.

kognate:
Data is recieved after the last byte is sent, so I can't close then.

On Mon, Mar 20, 2017, 1:28 PM German Attanasio <notifications@github.com>
wrote:

*@germanattanasio* commented on this pull request.
------------------------------

In watson_developer_cloud/speech_to_text_v1.py
<https://github.com/watson-developer-cloud/python-sdk/pull/177#discussion_r106964429>
:

> +                    self.sendMessage(json.dumps({'action': 'stop'}).encode('utf-8'))
+                    break
+                self.sendMessage(buf, isBinary=True)
+
+    def onMessage(self, payload, is_binary):
+        if not is_binary:
+            json_payload = json.loads(payload)
+            try:
+                self.messageCallback(json_payload)
+            except AttributeError:
+                # if we don't have a message callback, don't worry about it
+                pass
+            if 'results' in json_payload:
+                if len(json_payload['results']) > 0:
+                    if 'final' in json_payload['results'][0]:
+                        if json_payload['results'][0]['final']:

You should close it when there is no more data to be sent.

—
You are receiving this because you authored the thread.
Reply to this email directly, view it on GitHub
<https://github.com/watson-developer-cloud/python-sdk/pull/177#discussion_r106964429>,
or mute the thread

kognate:
@germanattanasio I think I've address all of your comments.  I'm still not happy with the twisted stuff, but it works on python 2.7 - 3.6

requirements.txt
```diff
@@ -1,3 +1,7 @@
 requests>=2.0,<3.0
 pysolr>=3.3,<4.0
 argparse>=1.3.0
+autobahn>=0.10.9
+Twisted>=13.2.0
+pyopenssl
```
germanattanasio: which version?

requirements.txt
```diff
@@ -1,3 +1,7 @@
...
+service-identity
```
germanattanasio: which version?

watson_developer_cloud/speech_to_text_v1.py
```diff
@@ -18,10 +18,133 @@
...
+class AudioFileProducer:
+    def __init__(self, proto):
+        self.proto = proto
+        self.started = False
+        self.paused = False
+        self.filedata = proto.filedatasource.pop()
+        self.fileobj = open(self.filedata.file_name, 'rb')
+        self.proto.metadata = self.filedata.file_metadata
+
+    def pauseProducing(self):
```
germanattanasio: should be `snake_case` and probably just `pause` since you will call the variable `producer` you can say

`producer.pause()`

watson_developer_cloud/speech_to_text_v1.py
```diff
@@ -18,10 +18,133 @@
...
+        self.paused = True
+
+    def resumeProducing(self):
```
germanattanasio: `snake_case`

watson_developer_cloud/speech_to_text_v1.py
```diff
@@ -18,10 +18,133 @@
...
+
+        while not self.paused:
+            buf = self.fileobj.read(1024)
+            if not buf:
+                as_json = json.dumps({"action": "stop"}).encode('utf-8')
+                self.proto.sendMessage(as_json, False)
+                break
+            self.proto.sendMessage(buf, True)
+
+    def stopProducing(self):
```
germanattanasio: `snake_case`

watson_developer_cloud/speech_to_text_v1.py
```diff
@@ -18,10 +18,133 @@
...
+
+class StreamingSpeechToTextProtocol(WebSocketClientProtocol):
+    filedatasource = deque()
+    params = {}
+    factory = None
+    ssl_context_factory = None
+    connects = 0
+
+    @classmethod
+    def buildFactory(cls, url, headers, params):
```
germanattanasio: `snake_case`

watson_developer_cloud/speech_to_text_v1.py
```diff
@@ -18,10 +18,133 @@
...
+    @classmethod
+    def start(cls):
+        reactor.run()
+
+    @classmethod
+    def stop(cls):
+        reactor.stop()
+
+    @classmethod
+    def addContentType(cls, content_type):
```
germanattanasio: `snake_case`

germanattanasio:
In the notebook the point (7) has code that's not indented

watson_developer_cloud/speech_to_text_v1.py
kognate: This is an interface I'm extending so the name is not changeable.

watson_developer_cloud/speech_to_text_v1.py
kognate: Same as above,  this is conforming to someone else's api

watson_developer_cloud/speech_to_text_v1.py
kognate: someone else's api

watson_developer_cloud/speech_to_text_v1.py
kognate: ok,  this code isn't part of someone elses api,  however, my argument is that since it has other camelCase methods that the methods of this class should all be consistent.

watson_developer_cloud/speech_to_text_v1.py
kognate: Same as above:  ok, this code isn't part of someone elses api, however, my argument is that since it has other camelCase methods that the methods of this class should all be consistent.

germanattanasio:
@kognate can you write the code that users will have to write to use this functionality?

I'm interested in:
 1. How the user sends the parameters
 1. How the user gets the `on_connection` message
 1. How the user gets the `on_listening` message
 1. How would it look when we want to send a stream of data
 1. How the user close the WebSocket connection
 1. How the user gets the `on_message` interim or final result 


I only need  1 example showcasing all the items above.

PR #178, opened 2017-03-21
Updates pending release of 0.25.2

kognate:
_2017-03-21_

* Fixed: Speech to text supports customization_id
* Fixed: Discovery now can create free instances


germanattanasio:
version should be updated with `bumpversion` 😋

PR #183, opened 2017-03-30
Adding better unicode support for Python 3

jsstylos:
Creating a pull request to get others to look at and test this before it's added. @kognate


jsstylos:
Looks like there are some issues on Python 3.3-3.5... I'll have to investigate more.

PR #189, opened 2017-04-04
Adds pyOpenSSL as a requirement

### LINKED ISSUE #188: Issue parsing new url to conversation_v1.py
We're using the Watson services in Frankfurt, Germany. The WCS in Germany has another gateway url. So we tried to parse a new url to the conversation_v1.py using url=https://gateway-fra.watsonplatform.net/conversation/api'. But we're always getting the following error. We are using Python 2.7.11.

Traceback (most recent call last):
  File "examples/flowtest.py", line 45, in <module>
    results = ft.runFlowTest(workspace_id=workspace_id, flow=flow, show_progress=True)
  File "/Users/yanick/Documents/conversation-test-suite-master-welcoming/ConversationTestKit/flowtest_v1.py", line 131, in runFlowTest
    r = self.ctk.message(workspace_id=workspace_id, message_input=payload, context=context)
  File "/Library/Python/2.7/site-packages/watson_developer_cloud/conversation_v1.py", line 181, in message
    json=data, accept_json=True)
  File "/Library/Python/2.7/site-packages/watson_developer_cloud/watson_developer_cloud_service.py", line 298, in request
    **kwargs)
  File "/Library/Python/2.7/site-packages/requests/api.py", line 56, in request
    return session.request(method=method, url=url, **kwargs)
  File "/Library/Python/2.7/site-packages/requests/sessions.py", line 488, in request
    resp = self.send(prep, **send_kwargs)
  File "/Library/Python/2.7/site-packages/requests/sessions.py", line 609, in send
    r = adapter.send(request, **kwargs)
  File "/Library/Python/2.7/site-packages/requests/adapters.py", line 497, in send
    raise SSLError(e, request=request)
requests.exceptions.SSLError: EOF occurred in violation of protocol (_ssl.c:590)

kognate:
This is added to address a bug in some users
OpenSSL.  Fixes #188


jsstylos:
Should this be in the `install_require` section of `setup.py` as well? Should it be a conditional requirement based on Python version?

kognate:
It should be in the `setup.py` but I don't think it should be conditional.  pyOpenSSL is probably required for all versions up to 3.5 and maybe beyond to solve this problem.

jsstylos:
Are we sure this is always a requirement? On my Mac, none of my pyenv environments have pyopenssl, including Python versions 2.7.12 and 3.5.0. Is it only on certain OSes?

kognate:
it's apparently a mac specific issue with accessing endpoints outside of US South.

PyOpenSSL is a pretty thin wrapper around system libs, and calls them if they exist.  

I've tried the code that breaks on MacOS on linux (Xenial, python 2.7) and it works just fine.  But on the mac,  it's reproducible (https://gist.github.com/kognate/4088374a65a93fb491789790c9ef0433).

jsstylos:
Huh, weird. I guess it's easy enough to just always require it. It would be nice to know why the non-us-south connections require it and us-south doesn't.

PR #193, opened 2017-04-13
Add Conversation intents, examples, and counterexamples

glennrfisher:
## Overview

This pull request adds support for the Conversation service's intent, examples, and counterexamples operations. Closes #181.

## Details

Most of the code was generated using `WatsonPythonCodegen`, although some manual changes were necessary (see 7219670ca53567f42d6658f518d1237b701dac71 for `conversation_v1.py` and 69e064932f896654b99bd82ccf5b1b701b00601d for `test_conversation_v1.py`).

A few notes that I'd like to draw your attention to:
- I used [YAPF][1] to automatically format the code. I also tried to make additional changes to adhere to the SDK's existing style. Let me know if there are any formatting changes I should make.
- There were naming conflicts between a path parameter and body variable for the `update_counterexample`, `update_intent`, and `update_example` operations. To resolve the conflict, I prefixed the body variables with `new_`.
- The unit tests follow the existing convention of verifying the URL and response object. Let me know if I should additional tests or assert statements.

## Tests

The unit tests pass on my machine:

```
(watson-python-sdk) Glenns-MacBook-Pro:test glennrfisher$ py.test test_conversation_v1.py
================================================= test session starts ==================================================
platform darwin -- Python 3.6.1, pytest-3.0.7, py-1.4.33, pluggy-0.4.0
rootdir: /Users/glennrfisher/Repositories/python-sdk, inifile:
plugins: cov-2.4.0
collected 21 items

test_conversation_v1.py .....................

============================================== 21 passed in 0.16 seconds ===============================================
```

The example also passes on my machine:

```
(watson-python-sdk) Glenns-MacBook-Pro:examples glennrfisher$ python conversation_v1.py
{
  "intents": [
    {
      "intent": "weather",
      "confidence": 0.9986543028393314
    }
  ],
  "entities": [],
  "input": {
    "text": "What's the weather like?"
  },
  "output": {
    "log_messages": [],
    "text": [
      "Hi. It looks like a nice drive today. What would you like me to do?"
    ],
    "nodes_visited": [
      "node_1_1467221909631"
    ]
  },
  "context": {
    "conversation_id": "31998782-8ef1-4002-a9e1-e8c4c4436d00",
    "system": {
      "dialog_stack": [
        {
          "dialog_node": "root"
        }
      ],
      "dialog_turn_counter": 1,
      "dialog_request_counter": 1,
      "branch_exited": true,
      "branch_exited_reason": "completed"
    },
    "defaultCounter": 0
  }
}
```

[...]

PR #195, opened 2017-04-20
Move `metadata` argument to avoid breaking changes

glennrfisher:
This pull request resolves a bug in #193. It moves the `metadata` argument in the `create_workspace` and `update_workspace` operations to be consistent with the existing handwritten parameter order. This fix prevents #193 from being a breaking change and requiring a major version release.


glennrfisher:
To check the parameter ordering, here's the [`create_workspace`][1] operation in the last version of the handwritten code.

[1]: https://github.com/watson-developer-cloud/python-sdk/blob/30c448275376cb23e9095dbe7fa3d3060cdb29b6/watson_developer_cloud/conversation_v1.py#L70

PR #196, opened 2017-04-21
Fix the build error for pull request # 195

mamoonraja:
Fix the build error for Python 2.7


kognate:
You've got a pylint problem, but other than that it looks good.

PR #198, opened 2017-05-01
Add tone_chat

glennrfisher:
This pull request adds support for the Tone Analyzer service's `tone_chat` operation.

- Existing `tone` tests pass.
- New `tone_chat` tests pass.
- New `tone_chat` example passes.


examples/tone_analyzer_v3.py
```diff
@@ -8,3 +8,6 @@
     version='2016-02-11')
 
 print(json.dumps(tone_analyzer.tone(text='I am very happy'), indent=2))
+
+utterances = [{'text': 'I am very happy', 'user': 'glenn'}]
```
jsstylos: Are you, Glenn? Are you??

watson_developer_cloud/tone_analyzer_v3.py
```diff
@@ -13,15 +13,25 @@
 # limitations under the License.
 
 """
-The v3 Tone Analyzer service
-(https://www.ibm.com/watson/developercloud/tone-analyzer.html)
+The IBM Watson™ Tone Analyzer Service uses linguistic analysis to detect
```
germanattanasio: I think the `™` is not an ASCII character. You need to add `# coding: utf-8` to the top of the file. It will be great if you can add it to the generator so that we don't run into this issue in the future.

glennrfisher:
@germanattanasio Good point! We wouldn't normally include the trademark symbol if writing the docstring by hand, so I updated the generator to remove it.

watson_developer_cloud/tone_analyzer_v3.py
glennrfisher: Good point! We wouldn't normally include the trademark symbol if writing the docstring by hand, so I updated the generator to remove it.

PR #201, opened 2017-05-05
Update Conversation for 04-21 release

glennrfisher:
This pull request uses watson-swagger-codegen to update Conversation for the 04/21 release.

Changes:
- Adds entity operations
- Adds synonym operations
- Adds values operations
- Adds log operation
- Adds an additional description to each docstring
- Initializes dictionaries with a literal
- Casts boolean query arguments to lowercase strings

Questions:
- Is the `logs` endpoint stable and ready to be public?
- In the future, should the generator explode arrays when the array element's model contains only one property? - For example, we explode the `CreateExample` body model for [`update_counterexample`][1], but not for [`update_workspace`][2].

[1]: https://github.com/watson-developer-cloud/python-sdk/blob/conversation/watson_developer_cloud/conversation_v1.py#L130
[2]: https://github.com/watson-developer-cloud/python-sdk/blob/conversation/watson_developer_cloud/conversation_v1.py#L1077


jsstylos:
Logs should be stable.

We probably don't want to simplify arrays of one-field objects into arrays of strings.

watson_developer_cloud/conversation_v1.py
```diff
@@ -89,21 +100,27 @@ def list_counterexamples(self,
...
         """
-        params = {'version': self.version}
-        params['page_limit'] = page_limit
-        params['include_count'] = include_count
-        params['sort'] = sort
-        params['cursor'] = cursor
+        params = {
+            'version': self.version,
+            'page_limit': page_limit,
+            'include_count': str(include_count).lower(),
```
glennrfisher: Great! I removed the code to cast boolean params to lowercase strings. I tested that with the examples and confirmed that the `_cleanup_param_values` was being called.

Should be ready to merge now. :)

germanattanasio:
`C:917, 0: Trailing whitespace (trailing-whitespace)`

glennrfisher:
Thanks for pointing that out, @germanattanasio! Looks like a `description` was missing from the Swagger definition, which caused trailing whitespace in a docstring. I added the missing description to the [Swagger definition][1] and updated [`conversation_v1.py`][2].

Running `pylint` also shows another warning: `W:571,18: Redefining built-in 'filter'`. I updated the generator to use [`log_filter`][3] instead. But let me know, @germanattanasio or @jsstylos, if there's a different name that you'd prefer.

[1]: https://github.ibm.com/Watson/developer-cloud--api-explorer/commit/ea87c57880bf6e25e84737aed75aa660b8e77663 
[2]: https://github.com/watson-developer-cloud/python-sdk/pull/201/commits/df51aa0d3e7f971c94ae5078295769278c16f1c9
[3]: https://github.com/watson-developer-cloud/python-sdk/pull/201/commits/0337d6cd3e1ee16ce2ab262ed9d88d4d6328fe70

jsstylos:
Hmm, `filter` is tricky, since it may appear in other functions that aren't `log`... I'm not immediately sure what the best name is.

glennrfisher:
The [current solution][1] is a bit hacky and definitely susceptible to issues. I think we'd like to move to vendor extensions before long, where we could write something like `"x-sdk-name": {"python": "log_filter"}` in the Swagger definition. We've talked about that briefly, but haven't gotten around to implementing it yet.

For now, we're only mapping `filter` to `log_filter` when generating the Conversation service. There aren't any other functions in Conversation that use `filter`, but that could definitely be an issue in the future.

[1]: https://github.ibm.com/MIL/watson-swagger-codegen/commit/0f7f7e18fadf8da8eda2c125a7fdd46868c37823

jsstylos:
To placate pylint about the redefinition warning, add: `,W0622` to the end of `.pylintrc`

glennrfisher:
Great, thanks for taking the time to look that up, @jsstylos! Just pushed a change to the `.pylintrc` file.

PR #205, opened 2017-05-18
Added update collection

IMpbt:
Added update collection to the discovery endpoint


germanattanasio:
@IMpbt Can you add a test case and fix travis?

PR #209, opened 2017-05-23
Add model management methods to Natural Language Understanding

g-may:
Successfully listed a custom model from my own service instance and deleted it with these methods.

### DIFF
--- a/watson_developer_cloud/natural_language_understanding_v1.py
+++ b/watson_developer_cloud/natural_language_understanding_v1.py
@@ -95,7 +95,7 @@ class NaturalLanguageUnderstandingV1(WatsonDeveloperCloudService):
         """
         Deletes a custom model
         :param model_id: The ID of the model to delete
-        :return: dict of analyzed text
+        :return: dict with status of model deletion
         """
         if model_id is None:
             raise ValueError("Missing parameter 'model_id'")
--- a/watson_developer_cloud/natural_language_understanding_v1.py
+++ b/watson_developer_cloud/natural_language_understanding_v1.py
@@ -91,15 +91,13 @@ class NaturalLanguageUnderstandingV1(WatsonDeveloperCloudService):
                             params={"version": self.version},
                             accept_json=True)
 
-    def deleteModel(self, model_id=None):
+    def deleteModel(self, model_id):
         """
         Deletes a custom model
         :param model_id: The ID of the model to delete
         :return: dict with status of model deletion
         """
-        if model_id is None:
-            raise ValueError("Missing parameter 'model_id'")
 
-        return self.request(method='DELETE', url='/v1/models/'+model_id,
+        return self.request(method='DELETE', url='/v1/models/{0}'.format(model_id),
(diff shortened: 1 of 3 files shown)


watson_developer_cloud/natural_language_understanding_v1.py
```diff
@@ -82,3 +82,24 @@ def analyze(self, features, text=None, url=None, html=None,
...
+        """
+        return self.request(method='GET', url='/v1/models',
+                            params={"version": self.version},
+                            accept_json=True)
+
+    def deleteModel(self, model_id=None):
+        """
+        Deletes a custom model
+        :param model_id: The ID of the model to delete
+        :return: dict of analyzed text
```
kognate: Model deletion really runs a dict of analyzed text?

watson_developer_cloud/natural_language_understanding_v1.py
```diff
@@ -82,3 +82,24 @@ def analyze(self, features, text=None, url=None, html=None,
...
+    def listModels(self):
+        """
+        Lists the custom models available for your service instance
+        :return: dict of available custom models
+        """
+        return self.request(method='GET', url='/v1/models',
+                            params={"version": self.version},
+                            accept_json=True)
+
+    def deleteModel(self, model_id=None):
```
kognate: You indicate that model_id can be None by making it's default value none.  I'd remove the default value and the model_id check later.    The service call will error on the None and by making it required the users knows that it should be there.

watson_developer_cloud/natural_language_understanding_v1.py
g-may: sure doesn't. thanks

watson_developer_cloud/natural_language_understanding_v1.py
```diff
@@ -82,3 +82,24 @@ def analyze(self, features, text=None, url=None, html=None,
...
+        """
+        Deletes a custom model
+        :param model_id: The ID of the model to delete
+        :return: dict with status of model deletion
+        """
+        if model_id is None:
+            raise ValueError("Missing parameter 'model_id'")
+
+        return self.request(method='DELETE', url='/v1/models/'+model_id,
```
kognate: Format strings   ` '/v1/models/{0}'.format(model_id)` are preferred over string concatenation.

watson_developer_cloud/natural_language_understanding_v1.py
g-may: I was just mirroring how the features parameter is checked by the SDK in the analyze endpoint. Does the new commit fix things?

watson_developer_cloud/natural_language_understanding_v1.py
g-may: fixed, thanks

kognate:
It does.

The difference is that params are optional in the analye method.  One of
them is required, and they are mutually exclusive.   In the listmodels
method the model_id is really required and isn't part of a set.

PR #210, opened 2017-05-26
Discovery update collection method and related test

IMpbt:


watson_developer_cloud/discovery_v1.py
```diff
@@ -226,6 +226,29 @@ def create_collection(self,
...
                             accept_json=True)
 
+    def update_collection(self,
+                          environment_id,
+                          collection_id,
+                          name,
+                          description="",
+                          configuration_id=None):
+
+        if configuration_id is None:
```
jsstylos: Sorry, just got to this pull request. Does this default configuration_id logic already live within the service? (When absent, doesn't the configuration_id default to the default configuration_id?)

jsstylos: If so we can just remove these lines, and having it set to `None` will cause it to be removed from the request which should allow the server to do the right thing automatically.

PR #226, opened 2017-06-28
Speech_to_Text_Websocket (Python 3.5) 

kstohr:
This adapts @daniel-bolanos [https://github.com/watson-developer-cloud/speech-to-text-websockets-python](speech-to-text-websockets-python) to python (3.5) based on the SpeechToTextV1  class of the Python SDK.  

- takes a *list* of paths to audio files ('file_input')
- returns a transcript for each file ['audio_file_name + .json']
- saves transcripts to a user designated directory ('dir_output')
- provides a log (as stdout or, for large audio files, as a log file saved to 'dir_output') 
- handles KeyboardInterrupt and ConnectionErrors with a kill function to prevent hanging
- latest service options ('speaker_labels' etc.) added to 'options'

Known issue: Autobahn's websocket 'reactor' can run only once per application.  Users may not be able to restart the reactor after a connection error or keyboard interrupt, for example, without restarting the interpreter. (This is why there are three test files, rather than one.)  Tried work arounds none worked.

Note: Could be back ported to Python 2 pretty easily. The queue.queue module was renamed, but all the modules are available etc. So, could be adapted work work with  2.7+

Needs review. 
Tested on python 3.5, mac os.x 10.11.6, chrome 59.0.3071.86


germanattanasio:
Hi @kstohr sorry for the delay. @ehdsouza ended up adding support for WebSocket using some of the ideas in this PR. I'm going to close this since the other PR was merge already.

We are planning on adding support for WebSockets in the upcoming release

PR #238, opened 2017-07-25
[discovery] Add training API wrappers

maniax89:


watson_developer_cloud/discovery_v1.py
```diff
@@ -402,3 +402,262 @@ def query(self, environment_id, collection_id, query_options):
...
+        :param natural_language_query:
+        :param filter:
+        :param examples:
+        :return: Training data for this collection found and returned.
+        """
+        format_string = '/v1/environments/{0}/collections/{1}/training_data'
+        url_string = format_string.format(environment_id, collection_id)
+        params = {'version': self.version}
+        data = {'natural_language_query': natural_language_query}
+        if filter:
```
jsstylos: Fields in `data` are automatically removed when None, so you don't need these checks.

watson_developer_cloud/discovery_v1.py
```diff
@@ -402,3 +402,262 @@ def query(self, environment_id, collection_id, query_options):
...
+        :param collection_id: the ID of your collection
+        :return: Code: 204, All training data removed
+        """
+        format_string = '/v1/environments/{0}/collections/{1}/training_data'
+        url_string = format_string.format(environment_id, collection_id)
+        params = {'version': self.version}
+
+        return self.request(method='DELETE',
+                            url=url_string,
+                            params=params)
```
jsstylos: Ideally the service itself would return JSON here, even if just an empty object or a message with the ID of the deleted resource. (For how the service is now, the SDK is fine.)

watson_developer_cloud/speech_to_text_v1.py
```diff
@@ -41,6 +41,8 @@ def recognize(self, audio, content_type, continuous=None, model=None,
         Returns the recognized text from the audio input
         """
         headers = {'content-type': content_type}
+        if continuous is None:
```
jsstylos: I think this may be due to changes in the underlying repo, but the changes to speech_to_text should be omitted -- these lines were removed to fix a deprecation bug.

watson_developer_cloud/discovery_v1.py
maniax89: yeah i noticed it was strange that it returns a `204` for no content so I just followed the API. I think your suggestion is valid and should be brought up in the API discussions to have it return something

watson_developer_cloud/speech_to_text_v1.py
maniax89: ok - maybe I should update the tests then instead of the code since the test expects to see `continuous=false` as a parameter in the query string.

watson_developer_cloud/discovery_v1.py
maniax89: 👍

maniax89:
actually I noticed that the SDK doesn't actually remove the `None` values from the request body if the body is JSON and keys are set to `None` - not sure if it is a limitation of Python 2.7 or what?

PR #239, opened 2017-07-26
Fix failing test in travis-ci add tox.ini for local testing

sdague:
The last commit changed a feature without updating the test, so is
failing in travis ci. This fixes that issue, as well as adding a
tox.ini file for easy local testing. Tox is a pretty standard test run
setup tool for python projects, and makes it easy to test against
multiple python versions simultaneously. The added one is boiler plate
for python 2.7, 3.4 -> 3.6, and has a pep8 rule.

Requests was removed from requirements-dev because it is also listed in requirements,
being in both breaks tox setting up the environment.

### DIFF
--- a/requirements-dev.txt
+++ b/requirements-dev.txt
@@ -8,7 +8,6 @@ pylint>=1.4.4
 coverage<4
 codecov>=1.6.3
 pytest-cov>=2.2.1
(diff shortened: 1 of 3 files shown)


jsstylos:
Thanks for the catches and fixes!

PR #248, opened 2017-08-09
Fix HTTP method for get_collection

rmoe:
This endpoint uses GET, not POST


germanattanasio:
@rmoe can you sign the CLA?

PR #250, opened 2017-08-11
generalize args for NLU

shuisman:
At the moment the args are fixed, which is not scalable. In line with other endpoints, this needs to be generalized.

### DIFF
--- a/watson_developer_cloud/natural_language_understanding_v1.py
+++ b/watson_developer_cloud/natural_language_understanding_v1.py
@@ -31,12 +31,10 @@ class NaturalLanguageUnderstandingV1(WatsonDeveloperCloudService):
                  url=default_url,
                  username=None,
                  password=None,
-                 use_vcap_services=True,
-                 x_watson_learning_opt_out=False):
+                 **kwargs):
         WatsonDeveloperCloudService.__init__(
             self, 'natural-language-understanding', url,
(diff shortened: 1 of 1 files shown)


germanattanasio:
@shuisman even if you make the arguments dynamic `** kwargs` they won't do anything because you are not using them in `WatsonDeveloperCloudService`

shuisman:
@germanattanasio, This is true. But at the moment the arguments are fixed. So for each new introduced functionality you have to add them here as well at the moment. With this change makes it future proof and in line with the other endpoints.
In example, `x_watson_learning_opt_out` is not available at NLU at the moment.

germanattanasio:
@shuisman new functionality doesn't actually change the constructor but If you want to do this you will have to do it for all the service. Right now you are only changing NLU

jsstylos:
We used to only support `x_watson_learning_opt_out` for Speech, but now I think more services are using it, so it's reasonable to use `**kwargs` to add it for the services that need it.

germanattanasio:
If we are going to support `**kwargs` I would prefer to support default headers. Like in the Node SDK

jsstylos:
We should add support for custom headers, and then switch the rest of the services to use **kwargs to support them

shuisman:
can this PR be merged please?

PR #269, opened 2017-09-29
Added Generated SDK files for new version of Tone

shubhanjan99:
Used `watson-swagger-codegen` to generate python files for latest version of Tone. Adds support for French, new API response for version parameter `2017-09-21`


germanattanasio:
The changes here are already in #276

PR #270, opened 2017-10-03
Add support for default_headers

mkistler:
This PR adds support for a `set_default_headers` in the base `WatsonDeveloperCloudService` class.  This is a more general and flexible solution for setting the `x-watson-learning-opt-out` header.  I left the `x_watson_learning_opt_out`parameter on the service constructor for backward compatibility -- we should consider removing that on the next major release.

### DIFF
--- a/test/test_alchemy_language_v1.py
+++ b/test/test_alchemy_language_v1.py
@@ -10,7 +10,6 @@ class TestAlchemyLanguageV1(TestCase):
         default_url = 'https://gateway-a.watsonplatform.net/calls'
         inited = watson_developer_cloud.AlchemyLanguageV1(url=default_url, api_key='boguskey',
                                                           x_watson_learning_opt_out=True)
-        #assert inited.x_watson_learning_opt_out
         assert inited.api_key == 'boguskey'
         assert inited.url == default_url
         inited.set_url(url="http://google.com")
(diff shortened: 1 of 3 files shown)


test/test_alchemy_language_v1.py
```diff
@@ -10,7 +10,7 @@ def test_api_key(self):
         default_url = 'https://gateway-a.watsonplatform.net/calls'
         inited = watson_developer_cloud.AlchemyLanguageV1(url=default_url, api_key='boguskey',
                                                           x_watson_learning_opt_out=True)
-        assert inited.x_watson_learning_opt_out
+        #assert inited.x_watson_learning_opt_out
```
glennrfisher: Can delete this line, since it's no longer needed.

test/test_conversation_v1.py
```diff
@@ -752,6 +753,8 @@ def test_message():
 
     assert message is not None
     assert responses.calls[0].request.url == message_url1
+    assert 'x-watson-learning-opt-out' in responses.calls[0].request.headers
+    assert responses.calls[0].request.headers['x-watson-learning-opt-out'] == 'true'
```
glennrfisher: Nice check on the response. More confidence-inspiring than the previous `assert inited.x_watson_learning_opt_out`.

test/test_alchemy_language_v1.py
mkistler: Done.

PR #276, opened 2017-10-10
Generated Python SDK for multiple services

mkistler:
Generated SDK and updated tests/examples for Conversation, Discovery, Language Translator, Natural Language Classifier, Natural Language Understanding, Personality Insights, Tone Analyzer, Visual Recognition.

This version of the SDK accepts either models or dicts for input parameters and produces dicts as method responses. Models for response classes are still generated and not pruned, so users can create a model from the returned dict if they so choose.

This version is not yet ready to merge but I've created the PR as a mechanism to get feedback.

### Breaking changes

While mostly compatible with the previously released Python SDK, there are some incompatible changes sprinkled through the services, including:

- Conversation:
  - message parameter "message_input" renamed to "input"
- Discovery:
  - add_document parameter "file_data" is removed. File contents are now passed with the file/filename parameters.
  - some methods are renamed, e.g., get_environments -> list_environments
- Language Translator
  - some methods are renamed, e.g. get_models -> list_models
- Natural Language Classifier
  - some methods are renamed, e.g. create -> create_classifier
- Natural Language Understanding
  - Dropped hand-written "Features" module in favor of generated Features model
- Tone Analyzer
  - tone parameter "tones" has been removed (new version date)
  - tone parameter "text" replaced by "tone_input"
  - tone parameter "content_type" default value changed from "text/plain" to "application/json"
  - tone parameters have been reordered
- Visual Recognition
  - classify parameters "images_url", "classifier_ids", "owners", and xxx replaced with "parameters"
  - Collections removed

Based on the above text, also added the Migration and Changelog files in the Wiki

PR #278, opened 2017-10-11
Query Multiple Collections 

rebecca-makar:
Performs a multiple collection query and returns the results from all collections that are sent in. Goes along with now having the ability through curl to query multiple collections and also query the notices.


ehdsouza:
Hi @rebecca-makar,

This is already being addressed in https://github.com/watson-developer-cloud/python-sdk/pull/276 as part of the new SDK release. It is named `federated_query`. Thanks for the effort.

PR #282, opened 2017-10-22
Use tox with Travis and some cleanup

germanattanasio:


germanattanasio:
@mkistler I need the `VCAP_SERVICES` values to update Travis to run the integration tests.

PR #286, opened 2017-10-24
[MAJOR RELEASE] Auto-generated code and other changes

germanattanasio:
Version 1.0 introduces automatically generated service classes for several services.

Services that are **not affected** in this release:

 * Speech to Text
 * Text to Speech
 * Retrieve and Rank
 * Document Conversion
 * Tradeoff Analytics

Services that are **affected** in this release:

 * Discovery
 * Conversation
 * Natural Language Understanding
 * Natural Language Classifier
 * Tone Analyzer
 * Language Translator
 * Personality Insights
 * Visual Recognition

Upon upgrading the SDK, follow the migration guide [here](https://github.com/watson-developer-cloud/python-sdk/wiki/Migration).


examples/conversation_tone_analyzer_integration/tone_conversation_integration.v1.py
```diff
@@ -29,11 +30,11 @@
...
-maintainToneHistoryInContext = True
+global_maintainToneHistoryInContext = True
 
 # Payload for the Watson Conversation Service
 # user input text required - replace "I am happy" with user input text.
-payload = {
+global_payload = {
     'workspace_id': workspace_id,
     'input': {
         'text': "I am happy"
```
mamoonraja: Line 47 under `invokeToneConversation` says:

`invokeToneConversation calls the the Tone Analyzer service to get the`

extra `the` is not intentional. right?

README.md
```diff
@@ -60,9 +59,11 @@ Copy your credentials from the **Service details** page. To find the the Service
 
 Tested on: Python from 2.7, 3.4, 3.5, and 3.6.
 
-## CHANGELOG
+## Changes for v4.0
+Version 4.0 focuses on the move to programmatically-generated code for many of the services. See the [changelog](https://github.com/watson-developer-cloud/python-sdk/wiki/Changelog) for the details.
```
glennrfisher: Should this be `v1.0` instead of `v4.0`?

examples/speech_to_text_v1.py
```diff
@@ -17,4 +18,4 @@
     print(json.dumps(speech_to_text.recognize(
         audio_file, content_type='audio/wav', timestamps=True,
         word_confidence=True),
-        indent=2))
+                     indent=2))
```
glennrfisher: It doesn't change the meaning, but I suspect this extra indentation wasn't intended.

examples/tone_analyzer_v3.py
```diff
@@ -1,45 +1,45 @@
...
-print(json.dumps(tone_analyzer.tone(text='I am very happy. It is a good day.'),
-                 indent=2))
+print(json.dumps(tone_analyzer.tone(tone_input='I am very happy. It is a good day.',
+                                    content_type="text/plain"), indent=2))
 
 print("\ntone() example 2:\n")
 with open(join(dirname(__file__),
                '../resources/tone-example.json')) as tone_json:
-    tone = tone_analyzer.tone(json.load(tone_json)['text'], 'emotion')
+    tone = tone_analyzer.tone(json.load(tone_json)['text'], "text/plain")
```
glennrfisher: Looks like the `tones` list was removed. Based on an email thread, though, I think there's a chance it will be reintroduced to maintain backwards compatibility. Not sure if y'all want to hold off on this release until there's a resolution to that discussion.

examples/visual_recognition_v3.py
```diff
@@ -34,21 +33,21 @@
 #
 # cars_positive_examples=image_file), indent=2))
 
-url_result = visual_recognition.classify(images_url=test_url)
+url_result = visual_recognition.classify(parameters=json.dumps({'url': test_url}))
 print(json.dumps(url_result, indent=2))
 
-faces_result = visual_recognition.detect_faces(images_url=test_url)
+faces_result = visual_recognition.detect_faces(parameters=json.dumps({'url': test_url}))
```
glennrfisher: It's unfortunate that the API for Visual Recognition expects a JSON string. Not sure what can be done to improve that without changing the service interface.

examples/visual_recognition_v3.py
```diff
@@ -34,21 +33,21 @@
...
 print(json.dumps(visual_recognition.list_classifiers(), indent=2))
 
-file_path = join(dirname(__file__), '../resources/text.png')
-with open(file_path, 'rb') as image_file:
-    text_results = visual_recognition.recognize_text(images_file=image_file)
-    print(json.dumps(text_results, indent=2))
+#file_path = join(dirname(__file__), '../resources/text.png')
+#with open(file_path, 'rb') as image_file:
+#    text_results = visual_recognition.recognize_text(images_file=image_file)
+#    print(json.dumps(text_results, indent=2))
```
glennrfisher: Was this intended to be commented?

watson_developer_cloud/discovery_v1.py
```diff
@@ -12,620 +14,4824 @@
...
+            query,
+            'natural_language_query':
+            natural_language_query,
+            'aggregation':
+            aggregation,
+            'count':
+            count,
+            'return_fields':
+            ",".join(return_fields)
+            if isinstance(return_fields, list) else return_fields,
```
glennrfisher: This looks awkward, although it seems to work okay. I suspect `yapf` formatted the code this way.

I made a quick test file, and it works as expected:

```python
# test.py
x = ['1','2','3']
params = {
    'list':
    ','.join(x)
    if True else 'hello, world'
}
print(params['list'])
```

```bash
> python test.py
1,2,3
```

examples/discovery_v1.py
```diff
@@ -1,13 +1,14 @@
 # coding: utf-8
+from __future__ import print_function
 import json
-import watson_developer_cloud
+from watson_developer_cloud import DiscoveryV1
 
-discovery = watson_developer_cloud.DiscoveryV1(
+discovery = DiscoveryV1(
     '2016-11-07',
```
mamoonraja: Using a latest version here would make more sense.

examples/tone_analyzer_v3.py
mkistler: Good eye @glennrfisher!  The `tones` parameter was reintroduced but is now `optional`, so it moved _after_ the `content_type` parameter.  That wouldn't really be backward compatible, but fortunately we don't have to worry about that for a new major release.  And happily, this example is still correct even with `tones` being reintroduced.

@ehdsouza The reordering of the parameters might be worth mentioning in the migration guide.

examples/visual_recognition_v3.py
mkistler: Yes -- the text recognition methods of VR are deprecated.

watson_developer_cloud/discovery_v1.py
mkistler: I noticed the same thing.  It turns out that if _any_ line in the object constructor has to break at the `:`, then yapf breaks all the lines consistently at the `:`.  It looks awful but it works.

watson_developer_cloud/language_translator_v2.py
```diff
@@ -11,78 +13,812 @@
...
-                 'monolingual_corpus': monolingual_corpus}
-        return self.request(method='POST', url='/v2/models', params=params,
-                            files=files, accept_json=True)
+                     monolingual_corpus=None,
+                     forced_glossary_content_type=None,
+                     forced_glossary_filename=None,
+                     parallel_corpus_content_type=None,
+                     parallel_corpus_filename=None,
+                     monolingual_corpus_content_type=None,
+                     monolingual_corpus_filename=None):
```
mkistler: I think we probably do not want the `content_types` or `filename` parameters here.

examples/speech_to_text_v1.py
ehdsouza: This was a pylint complain, so added those indentation

watson_developer_cloud/language_translator_v2.py
mkistler: I see why this happened -- a typo in the annotations in the LT swagger.

examples/visual_recognition_v3.py
ehdsouza: :(

watson_developer_cloud/natural_language_classifier_v1.py
```diff
@@ -11,59 +13,459 @@
...
+            method='POST',
+            url='/v1/classifiers/{0}/classify'.format(classifier_id),
+            json=data,
+            accept_json=True)
+        return response
+
+    def create_classifier(self,
+                          metadata,
+                          training_data,
+                          training_metadata_filename=None,
```
mkistler: This parameter should be named `metadata_filename`.

examples/tone_analyzer_v3.py
ehdsouza: Updated the migration guide.

examples/discovery_v1.py
germanattanasio: Use `2017-10-16`

watson_developer_cloud/conversation_v1.py
```diff
@@ -11,108 +13,200 @@
...
+               the most recent version before that date. Note that you should
+               not programmatically specify the current date at runtime, in
+               case the API has been updated since your application's release.
+               Instead, specify a version date that is compatible with your
+               application, and don't change it until your application is
+               ready for a later version.
 
-    def __init__(self, version, url=default_url, **kwargs):
-        WatsonDeveloperCloudService.__init__(self, 'conversation', url,
-                                             **kwargs)
```
jsstylos: Should continue to have the logic of `**kwargs` to support `x_watson_learning_opt_out`, etc.

```    def __init__(self, version, url=default_url, **kwargs):
        WatsonDeveloperCloudService.__init__(self, 'conversation', url,
                                             **kwargs)```

File: watson_developer_cloud/conversation_v1.py [hunk 12]
```diff
@@ -11,108 +13,200 @@
...
+        :param list[CreateEntity] entities: An array of objects defining the entities for the workspace.
+        :param list[CreateDialogNode] dialog_nodes: An array of objects defining the nodes in the workspace dialog.
+        :param list[CreateCounterexample] counterexamples: An array of objects defining input examples that have been marked as irrelevant input.
+        :param object metadata: Any metadata related to the workspace.
+        :param bool learning_opt_out: Whether training data from the workspace can be used by IBM for general service improvements. `true` indicates that workspace training data is not to be used.
+        :return: A `dict` containing the `Workspace` response.
+        :rtype: dict
         """
+        if intents is not None:
+            intents = [
```
jsstylos: This is an implementation code readability issue, not an API issue, but can we extract this logic out so that it happens with a single call rather than seeing a bunch of lines of code for each parameter? We have similar logic in the base class right now for automatically converting None values, etc.

watson_developer_cloud/conversation_v1.py
```diff
@@ -312,55 +529,81 @@ def create_example(self, workspace_id, intent, text):
...
+            raise ValueError('text must be provided')
         params = {'version': self.version}
-        return self.request(
+        self.request(
             method='DELETE',
             url='/v1/workspaces/{0}/intents/{1}/examples/{2}'.format(
                 workspace_id, intent, text),
             params=params,
             accept_json=True)
+        return None
```
jsstylos: We should return the empty object, not `None`, so that users can always try to display the response without it blowing up.

watson_developer_cloud/visual_recognition_v3.py
```diff
@@ -12,365 +14,1331 @@
...
         """
 
-        WatsonDeveloperCloudService.__init__(self, 'watson_vision_combined',
-                                             url, **kwargs)
+        WatsonService.__init__(
+            self,
+            vcap_services_name='watson_vision_combined',
+            url=url,
+            api_key=api_key,
+            use_vcap_services=True)
```
jsstylos: Part of the `**kwargs` issue, but we should still let users set this false when needed, not just hardcode it as true

watson_developer_cloud/conversation_v1.py
mkistler: You can do `x_watson_learnign_opt_out` with the default headers.  This was recently added.

@ehdsouza We should make sure this is in the migration guide.

mkistler: Also, `**kwargs` is no longer needed for `use_vcap_services`, because it defaults to `True` -- there’s never a case where you want to specify `False`.  Credentials passed in always override vcap services.  If you want vcap services, don’t pass credentials.

watson_developer_cloud/conversation_v1.py
mkistler: Well, it only wraps to multiple lines because the 80 char line length restriction, but yes, we can look at creating a utility method for this.

watson_developer_cloud/conversation_v1.py
mkistler: Happily, there is no problem with trying to print `None`.  Python handles it just fine without any exceptions:
```
[mkistler@mkistlers-MacBook-Pro] ~>python
Python 2.7.12 (default, Oct 11 2016, 05:20:59) 
[GCC 4.2.1 Compatible Apple LLVM 8.0.0 (clang-800.0.38)] on darwin
Type "help", "copyright", "credits" or "license" for more information.
>>> import json
>>> response = None
>>> try:
...     print(json.dumps(response, indent=2)) 
... except:
...     print("exception!")
... 
null
>>> 
```
If some other version of Python did not handle this as gracefully, that is easily fixed:
```
>>> print(json.dumps(response or {}, indent=2)) 
{}
>>> 
```

watson_developer_cloud/conversation_v1.py
mkistler: How about this:  Create this utility method:
```
def _convert_model(modelOrDict):
    if hasattr(x, "_to_dict"):
        return x._to_dict()
    return x
```
So that the code in the service can now look like:
```
        if intents is not None:
            intents = [ _convert_model(x) for x in intents ]
```

germanattanasio: You can make it better by detecting when it's an array and doing the for there

mkistler: @german I'm not sure I follow.  Are you saying that we should not assume that `intents` is an array -- as this code seems to do -- and instead we should wrap this code in a check that `intents` is an array?

germanattanasio: I was thinking that we can simplify the `_convert_model` code and make it work for all the cases. With some introspection we can see if is an array or not.

germanattanasio:
selfie-0

PR #297, opened 2017-11-05
Create custom exception for API errors containing the status code

### LINKED ISSUE #279: Exceptions don't have enough information for users
As a Developer _I want_ to have `code`, `message`, `description` whenever a Watson service raises an exception _so that_ I can write code that reacts different depending on the exception.

# Context
It seems like the Python exceptions this SDK is throwing are not very useful for developers:
> **Bruce Adams:**
> I do not have a strong opinion about exception classes. I do find it irritating that `WatsonException` takes a structured error response from as service call and flattens it into a string as the exception args. When I want to make decisions based on the error code (my code needs to react differently to 429 compared to a 500), I have to pick the code out of the string.

More info:
 * https://stackoverflow.com/questions/256222/which-exception-should-i-raise-on-bad-illegal-argument-combinations-in-python
 * https://www.youtube.com/watch?v=o9pEzgHorH0
## Acceptance Criteria
- [ ] Switch to use `ValueError` when there is a missing or illegal argument
- [ ] Parse the error parameters and make them available to users (`code`, `description`)

mkistler:
This PR fixes #279 by adding a new custom exception class for API errors with explicit fields for the status code and extra info.

### DIFF
--- a/watson_developer_cloud/watson_service.py
+++ b/watson_developer_cloud/watson_service.py
@@ -62,6 +62,7 @@ class WatsonApiException(WatsonException):
     def __init__(self, code, message, info=None):
         # Call the base class constructor with the parameters it needs
         super(WatsonApiException, self).__init__(message)
+        self.message = message
         self.code = code
         self.info = info
 
@@ -232,7 +233,7 @@ class WatsonService(object):
                 error_message = error_json['statusInfo']
             return error_message
         except:
-            return (response.text or error_message)
+            return response.text or error_message
 
 
     @staticmethod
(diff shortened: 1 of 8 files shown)


watson_developer_cloud/watson_service.py
```diff
@@ -45,9 +45,30 @@ def load_from_vcap_services(service_name):
...
+        super(WatsonApiException, self).__init__(message)
+        self.code = code
+        self.info = info
+
+    def __str__(self):
+        return 'Error: ' + self.message + ', Code: ' + str(self.code)
+
+
 class WatsonInvalidArgument(WatsonException):
     pass
```
germanattanasio: Should `_cleanup_param_value`, `_cleanup_param_values` and all the other clean methods be `@ staticmethod`?

germanattanasio:
I think you are missing an import on `test_conversation_v1`

PR #309, opened 2017-11-25
Add customization_weight parameter to recognize() for stt

### LINKED ISSUE #306: Support for multi-part audio, and customization_weight parameters in the Python SDK for STT
Hi,
Unless I am missing something, I don't see how to use the STT python libraries to process multipart audio files (cURL command lets you do this)... 
Also, is there a way to pass content headers, and the parameter customization_weight etc to the recognize call using the Python SDK?

ammardodin:
Fixes #306

### DIFF
--- a/watson_developer_cloud/speech_to_text_v1.py
+++ b/watson_developer_cloud/speech_to_text_v1.py
@@ -35,7 +35,8 @@ class SpeechToTextV1(WatsonService):
                   word_confidence=None, timestamps=None, interim_results=None,
                   profanity_filter=None,
                   smart_formatting=None,
-                  speaker_labels=None):
+                  speaker_labels=None,
+                  customization_weight=None):
         """
         Returns the recognized text from the audio input
         """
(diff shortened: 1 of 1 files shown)


watson_developer_cloud/speech_to_text_v1.py
```diff
@@ -55,6 +56,10 @@ def recognize(self, audio, content_type, continuous=None, model=None,
                   'smart_formatting': smart_formatting,
                   'speaker_labels': speaker_labels}
 
+        if (params['customization_id'] is not None and 
+            customization_weight is not None):
+            params['customization_weight'] = customization_weight
+
```
mkistler: I think you can simply add
```
    'customization_weight': customization_weight,
```
into the params object above.  The check for `customization_weight is not None` isn't needed because the service will strip it out, and the check for 'customization_id' _shouldn't_ be needed because the service should simply ignore it when there is no customization_id.

ammardodin: Ah, touche. I wasn't counting on the service to do that 😛 , but that's good to know !

PR #323, opened 2017-12-06
Add utility method to convert lists to csv.

mkistler:
This PR adds a helper method to the base service class to handle conversion of lists to comma-separated strings.  This allows service methods to use more concise (and PEP8-friendly) logic for list parameters.


ammardodin:
Looks good 👍

Issue #327, opened 2017-12-11
Feature Request: Allow specification of connection timeout


breathe:
Would be nice if we could pass a parameter through to the 'requests' library to control connection timeout.

germanattanasio:
This is a good idea. It will also help if we can specify a proxy.

germanattanasio:
@breathe I think it will take some time because people are on vacation. Feel free to open a PR if you have the time.

breathe:
@germanattanasio I can submit a PR -- I'm thinking:

timeout as an additional optional argument to DiscoveryV1 __init__
```
class DiscoveryV1(WatsonDeveloperCloudService):
    """Client for Discovery service"""

    def __init__(self,
                 version,
                 url=default_url,
                 username=None,
                 password=None,
                 use_vcap_services=True,
                 timeout=None):
``` 

And then adding timeout parameter within DiscoveryV1.request method:
```
        response = requests.request(method=method, url=full_url,
                                    cookies=self.jar, auth=auth,
                                    headers=headers,
                                    params=params, data=data, files=files,
                                    timeout=self.timeout
                                    **kwargs)
```

Would let me get rid of this ugly monkeypatch:

```
# EVIL: MONKEY PATCH THE REQUESTS LIBRARY TO ADD TIMEOUT TO ALL requests.request() calls
originalRequestFunction = requests.request


def monkeyPatchedRequest(*args, **kwords):
    return originalRequestFunction(*args, **kwords, timeout=(5, 5))


requests.request = monkeyPatchedRequest
# END EVIL
```

germanattanasio:
I think that having `timeout` in the `init()` method is enough. Should the name be `timeout`? should we wrap `timeout` and other parameters like `proxies` in a dictionary. I'm thinking of something like 
`http_config`. 🤔 

```py
def __init__(self,
                 version,
                 url=default_url,
                 username=None,
                 password=None,
                 use_vcap_services=True,
                 http_config=None):
```

breathe:
http_config as a dictionary might be closer to the right thing.  I'll update the PR

ehdsouza:
@breathe thanks for suggesting this. It has been handled in PR: https://github.com/watson-developer-cloud/python-sdk/pull/340

PR #328, opened 2017-12-11
Allow passing dictionary of http client parameters to DiscoveryV1 api constructor

breathe:
Fix for #327


germanattanasio:
@jsstylos can you take a look at this? I think this is not a breaking change so we could add it to the generator.

jsstylos:
This looks good, we will need to add it in the generator due to our new SDK generation process, but that also means we will be able to add it to all of the services, not just Discovery.

watson_developer_cloud/discovery_v1.py
```diff
@@ -43,7 +43,7 @@ class DiscoveryV1(WatsonService):
     VERSION_DATE_2017_06_25 = '2017-06-25'
     VERSION_DATE_2016_12_01 = '2016-12-01'
 
-    def __init__(self, version, url=default_url, username=None, password=None):
+    def __init__(self, version, url=default_url, username=None, password=None, http_config={}):
```
mkistler: I'd like to suggest an alternate solution -- just add a method to the `watson_service` class for setting the http_config.  This is probably needed anyway, and once we have it, we don't really need to add an `http_config` parameter to all the service constructors, or change the generator -- it's a much simpler solution.

jsstylos: That sounds like a good idea.

PR #332, opened 2017-12-12
Generate Text to Speech

ammardodin:
**Summary**

* Generated `Text to Speech` from the latest swagger
* Added missing unit tests for `get_voice`
* Updated method names in unit tests
* Integration tests
* Updated examples as per new .env file


watson_developer_cloud/text_to_speech_v1.py
```diff
@@ -12,123 +14,1198 @@
...
+        includes invalid query parameters, the service returns a `Warnings` response
+        header that provides messages about the invalid parameters. The warning includes a
+        descriptive message and a list of invalid argument strings. For example, a message
+        such as `\"Unknown arguments:\"` or `\"Unknown url query arguments:\"` followed by
+        a list of the form `\"invalid_arg_1, invalid_arg_2.\"` The request succeeds
+        despite the warnings.
+
+        :param str text: The text to synthesize.
+        :param str accept: The requested audio format (MIME type) of the audio. You can use this header or the `accept` query parameter to specify the audio format. (For the `audio/l16` format, you can optionally specify `endianness=big-endian` or `endianness=little-endian`; the default is little endian.).
+        :param str accept2: The requested audio format (MIME type) of the audio. You can use this query parameter or the `Accept` header to specify the audio format. (For the `audio/l16` format, you can optionally specify `endianness=big-endian` or `endianness=little-endian`; the default is little endian.).
```
mkistler: We should ignore the `accept` query parameter -- it duplicates the function of the `accept` header parameter, and the header parameter is the better approach.

watson_developer_cloud/text_to_speech_v1.py
```diff
@@ -12,123 +14,1198 @@
...
+        """
+        if text is None:
+            raise ValueError('text must be provided')
         params = {
-            'language': language
-        }
-        return self.request(method='GET', url='/v1/customizations',
-                            params=params, accept_json=True)
-
-    def get_customization(self, customization_id):
```
mkistler: It will break compatibility to remove these methods, and I think that's a problem.  Why not leave them?  I don't think they conflict with any of the generated methods.  Maybe we can mark them as deprecated, and add comments recommending the new method names.  That will set us up to remove them in the next major release.

mkistler:
We can use Python's `warnings` package to add deprecation warnings to existing methods that have different names in the generated SDK.

http://www.arruda.blog.br/programacao/python-use-warnings/

watson_developer_cloud/text_to_speech_v1.py
```diff
@@ -12,123 +14,1162 @@
...
+    def __ne__(self, other):
+        """Return `true` when self and other are not equal, false otherwise."""
+        return not self == other
+
+
+class Translation(object):
+    """
+    Translation.
+
+    :attr str translation: The phonetic or sounds-like translation for the word. A phonetic translation is based on the SSML format for representing the phonetic string of a word either as an IPA translation or as an IBM SPR translation. A sounds-like is one or more words that, when combined, sound like the word.
```
mkistler: Translation.translation is going to cause problems for C#.

watson_developer_cloud/text_to_speech_v1.py
```diff
@@ -12,123 +14,1162 @@
...
+    def __ne__(self, other):
+        """Return `true` when self and other are not equal, false otherwise."""
+        return not self == other
+
+
+class VoiceModels(object):
+    """
+    VoiceModels.
+
+    :attr list[VoiceModel] customizations: An array of `VoiceModel` objects that provides information about each available custom voice model. The array is empty if the requesting service credentials own no custom voice models (if no language is specified) or own no custom voice models for the specified language.
```
mkistler: Seems like we want this to be named `voiceModels` and not `customizations`.  I think this would be as simple as adding an `x-alternate-name` annotation to the swagger.

PR #335, opened 2017-12-15
Generate Speech to Text

ammardodin:
* Generated Speech to Text from latest Swagger.
* Manual tweaks to `add_word` and `add_words`
* Update old tests to use new signatures
* Add tests for acoustic customizations, recognitions, callbacks, custom audio resources
* Add integration tests


germanattanasio:
Some of the methods require a `content-type` even that they only support `application/json`. This is a problem.

PR #340, opened 2018-01-02
Set http client config as a separate method

ehdsouza:
based on the review comments in https://github.com/watson-developer-cloud/python-sdk/pull/328/files, we would want a method to set the client http configs so that no changes in the generator are needed


breathe:
@ehdsouza Any idea when this will make it into a public release?  I'm not finding this functionality in 1.0.2.

germanattanasio:
@breathe Friday Feb 2nd

PR #376, opened 2018-02-23
Speech to Text using web sockets 

ehdsouza:
It takes in an audio file and outputs interim results, hypothesis based on what the listener is interested in.

Following is the implementation details:
* Pass in a callback instance of type `RecognizeCallback` 
* In onOpen, it passed it the open action and options
* In onMessage, the appropriate callbacks are called.
* When the end of the audio is reached, a close action is sent and the reactor is closed which terminates the session

Tested on python 2.7 and 3.6.4
Written an example which works successfully
```
class MyRecognizeCallback(RecognizeAbstractCallback):
    def __init__(self):
        pass

    def on_transcription(self, transcript):
        print(transcript)

    def on_connected(self):
        print('Connection was successful')

    def on_error(self):
        print('Error received: {}'.format(error))

    def on_inactivity_timeout(self):
        print('Timeout!')

    def on_listening(self):
        print('Service is listening')

    def on_transcription_complete(self):
        print('Transcription completed')

    def on_hypothesis(self, hypothesis):
        print(hypothesis)

mycallback = MyRecognizeCallback()
with open(join(dirname(__file__), '../resources/speech.wav'),
          'rb') as audio_file:
    speech_to_text.recognize_with_websocket(
        audio=audio_file, recognize_callback=mycallback)
```

Another example has a microphone input for speech to text using web socket


examples/speech_to_text_v1.py
```diff
@@ -21,3 +23,35 @@
...
+# Example using websockets
+class MyRecognizeCallback(RecognizeAbstractCallback):
+    def __init__(self):
+        pass
+
+    def on_transcription(self, transcript):
+        print(transcript)
+
+    def on_connected(self):
+        print('Connection was successful')
```
germanattanasio: use the method name as the string
`print('on_connected()')`

requirements.txt
```diff
@@ -1,2 +1,7 @@
 requests>=2.0,<3.0
 python_dateutil>=2.5.3
+autobahn>=0.10.9
+Twisted>=13.2.0
+pyOpenSSL>=16.2.0
```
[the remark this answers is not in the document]
ehdsouza: yes, needed here: `from twisted.internet import ssl`

watson_developer_cloud/websocket/recognize_abstract_callback.py
```diff
@@ -0,0 +1,68 @@
...
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+from abc import abstractmethod, ABCMeta
+
+class RecognizeAbstractCallback:
+  __metaclass__ = ABCMeta
+
+  """
```
germanattanasio: I think the documentation goes into the method body not the header

watson_developer_cloud/websocket/speech_to_text_websocket_listener.py
```diff
@@ -0,0 +1,174 @@
...
+
+        if factory.isSecure:
+            contextFactory = ssl.ClientContextFactory()
+        else:
+            contextFactory = None
+        connectWS(factory, contextFactory)
+
+        reactor.run()
+
+    class WSInterfaceProtocol(WebSocketClientProtocol):
```
germanattanasio: Use `WebSocketClient` as name

watson_developer_cloud/websocket/speech_to_text_websocket_listener.py
```diff
@@ -0,0 +1,174 @@
...
+            self.factory.endReactor()
+
+    class WSInterfaceFactory(WebSocketClientFactory):
+        def __init__(self, audio, options, callback, url=None, headers=None):
+            WebSocketClientFactory.__init__(self, url=url, headers=headers)
+            self.audio = audio
+            self.options = options
+            self.callback = callback
+
+            self.openHandshakeTimeout = 6
```
germanattanasio: What's 6? is it seconds? milliseconds?.
Please use a constant like `SIX_SECONDS`

watson_developer_cloud/websocket/speech_to_text_websocket_listener.py
```diff
@@ -0,0 +1,174 @@
...
+            self.closeHandshakeTimeout = 6
+
+        def endReactor(self):
+            reactor.stop()
+
+        # this function gets called every time connectWS is called (once per WebSocket connection/session)
+        def buildProtocol(self, addr):
+            proto = RecognizeListener.WSInterfaceProtocol(
```
germanattanasio: `buildProtocol` or `buildWebSocketClient` ?

watson_developer_cloud/websocket/speech_to_text_websocket_listener.py
```diff
@@ -0,0 +1,174 @@
...
+
+                elif self.listening_messages == 2:
+                    # close the connection
+                    self.sendMessage(self.build_close_message())
+                    self.callback.on_transcription_complete()
+                    self.sendClose(1000)
+
+            # if in streaming
+            elif 'results' in json_object or 'speaker_labels' in json_object:
+                hypothesis = ""
```
germanattanasio: use `'` not `"`

watson_developer_cloud/websocket/speech_to_text_websocket_listener.py
```diff
@@ -0,0 +1,174 @@
...
+            elif 'state' in json_object:
+                self.listening_messages += 1
+                if self.listening_messages == 1:
+                    self.callback.on_listening()
+
+                elif self.listening_messages == 2:
+                    # close the connection
+                    self.sendMessage(self.build_close_message())
+                    self.callback.on_transcription_complete()
+                    self.sendClose(1000)
```
germanattanasio: use a constant `1000` should be `CLOSE_SIGNAL` or something like that

watson_developer_cloud/websocket/speech_to_text_websocket_listener.py
```diff
@@ -0,0 +1,174 @@
...
+                # options.inactivity_timeout
+                error = json_object['error']
+                if error.startswith(self.TIMEOUT_PREFIX):
+                    self.callback.on_inactivity_timeout()
+                else:
+                    self.callback.on_error()
+
+            # if uninitialized, receive the initialization response from the server
+            elif 'state' in json_object:
+                self.listening_messages += 1
```
germanattanasio: You will only get up to 1 listening message so I think you should use a boolean. What you want to know is if it's the first message which means it's listening or if it's the second message which says it's done processing the audio file.

watson_developer_cloud/websocket/speech_to_text_websocket_listener.py
ehdsouza: This function name is predefined in the `WebSocketClientFactory`

examples/speech_to_text_v1.py
ehdsouza: This is just an example code. A user can put in any statements he/she likes ;)

watson_developer_cloud/websocket/speech_to_text_websocket_listener.py
```diff
@@ -0,0 +1,174 @@
...
+            self.callback = callback
+            self.SIX_SECONDS = 6
+            self.openHandshakeTimeout = self.SIX_SECONDS
+            self.closeHandshakeTimeout = self.SIX_SECONDS
+
+        def endReactor(self):
+            reactor.stop()
+
+        # this function gets called every time connectWS is called (once per WebSocket connection/session)
+        def buildProtocol(self, addr):
```
germanattanasio: return the class directly
```py
        def buildProtocol(self, addr):
           return RecognizeListener.WebSocketClient(
                self, self.audio, self.options, self.callback)
```

watson_developer_cloud/websocket/speech_to_text_websocket_listener.py
```diff
@@ -0,0 +1,174 @@
...
+                    self.sendMessage(b'', isBinary=True)
+
+            if (self.bytes_sent + self.ONE_KB >= len(data)):
+                if (len(data) > self.bytes_sent):
+                    send_chunk(data[self.bytes_sent:len(data)], True)
+                    return
+
+            send_chunk(data[self.bytes_sent:self.bytes_sent + self.ONE_KB])
+            self.factory.reactor.callLater(self.TEN_MILLISECONDS, self.send_audio, data=data)
+            return
```
germanattanasio: Line 83 isn't necessary

watson_developer_cloud/websocket/speech_to_text_websocket_listener.py
```diff
@@ -0,0 +1,174 @@
...
+
+    class WebSocketClient(WebSocketClientProtocol):
+        def __init__(self, factory, audio, options, callback):
+            self.factory = factory
+            self.audio = audio
+            self.options = options
+            self.callback = callback
+            self.isListening = False
+            self.bytes_sent = 0
+            self.ONE_KB = 1000  # in bytes
```
germanattanasio: Use constants not class variables.

examples/speech_to_text_v1.py
```diff
@@ -21,3 +23,35 @@
                 timestamps=True,
                 word_confidence=True),
             indent=2))
+
+# Example using websockets
+class MyRecognizeCallback(RecognizeAbstractCallback):
```
germanattanasio: Do users need to override each method in `RecognizeAbstractCallback`? If that's the case let's make `RecognizeAbstractCallback` a concrete class and leave each method as 

```py
def method(self):
    pass
```

watson_developer_cloud/websocket/speech_to_text_websocket_listener.py
```diff
@@ -0,0 +1,174 @@
...
+                transcript['transcript'] = alternative['transcript']
+                transcripts.append(transcript)
+            return transcripts
+
+        def onConnect(self, response):
+            log.msg("onConnect, server connected: {}".format(response.peer))
+            self.callback.on_connected()
+
+        def onOpen(self):
+            log.msg("onOpen, Websocket connection open")
```
germanattanasio: why are you logging messages using the twisted log? If we want to add logging to the SDK we should use a standard python logging library instead.

watson_developer_cloud/websocket/speech_to_text_websocket_listener.py
```diff
@@ -0,0 +1,174 @@
...
+                    self.sendMessage(self.build_close_message())
+                    self.callback.on_transcription_complete()
+                    self.sendClose(self.CLOSE_SIGNAL)
+
+            # if in streaming
+            elif 'results' in json_object or 'speaker_labels' in json_object:
+                hypothesis = ''
+                # empty hypothesis
+                if len(json_object['results']) == 0:
+                    print('empty hypothesis!')
```
germanattanasio: Don't use `print`

watson_developer_cloud/websocket/speech_to_text_websocket_listener.py
```diff
@@ -0,0 +1,174 @@
...
+            if 'error' in json_object:
+                log.msg('Error : {}'.format(json_object['error']))
+                # Only call on_error() if a real error occurred. The STT service sends
+                # // {"error" : "No speech detected for 5s"} for valid timeouts, configured by
+                # options.inactivity_timeout
+                error = json_object['error']
+                if error.startswith(self.TIMEOUT_PREFIX):
+                    self.callback.on_inactivity_timeout()
+                else:
+                    self.callback.on_error()
```
germanattanasio: you need to send the error
```py
self.callback.on_error(error)
```

same with `inactivity_timeout()`

ehdsouza: Thanks for this :)

PR #394, opened 2018-03-07
Regenerate visual recognition

ehdsouza:


germanattanasio:
@ehdsouza we need to make sure we move all the fixes to pylint to the generator so that p

PR #398, opened 2018-03-08
Merge next to develop prior to 1.1.0 release

germanattanasio:
Changelog: https://github.com/watson-developer-cloud/python-sdk/wiki/Changelog#version-110


watson_developer_cloud/__init__.py
```diff
@@ -16,24 +17,17 @@
 from .watson_service import WatsonException
 from .watson_service import WatsonApiException
 from .watson_service import WatsonInvalidArgument
-from .alchemy_data_news_v1 import AlchemyDataNewsV1
-from .alchemy_language_v1 import AlchemyLanguageV1
-from .alchemy_vision_v1 import AlchemyVisionV1
```
ammardodin: Instead of completely removing these, we should use deprecation messages like we do the `node-sdk` so that users who try to instantiate those classes get a deprecation message instead of just an error.

ehdsouza: For this we will have to add the class files back and add the deprecation warnings in the class files

germanattanasio:
@ammardodin I was talking to Erika and there is no easy way to do this without adding back the deprecated files

Issue #410, opened 2018-03-16
conversation client .message() fails due to check for conversation_id


scottdangelo:
#### Expected behavior
Check in:
`watson_developer_cloud/conversation_v1.py", line 1716, in _from_dict` should pass

#### Actual behavior
This throws an exception because it is accessing the `conversation_id` field without using 
`['context']['conversation_id']`

#### Steps to reproduce the problem
use the .message() API with a message consisting of:
```
{'intents': [], 'entities': [], 'input': {}, 'output': {'text': ['Hello! and Welcome to Watson Online Store.'], 'nodes_visited': ['Welcome'], 'log_messages': []}, 'context': {'conversation_id': 'c1dfce32-3e10-446a-8bd5-8455f8ebe24e', 'system': {'dialog_stack': [{'dialog_node': 'Welcome'}], 'dialog_turn_counter': 1, 'dialog_request_counter': 1, '_node_output_map': {'Welcome': [0]}}, 'get_input': 'yes', 'input_text': '', 'shopping_cart': 'list'}, 'type': 'customer', 'email': 'scott.dangelo@ibm.com', 'first_name': 'Scott', 'last_name': 'DAngelo', 'shopping_cart': []}
```

#### python sdk version
Current version, but there is no `latest_version`
#### python version
Python 3.6.4

scottdangelo:
More details here:
https://github.com/IBM/watson-online-store/issues/134

scottdangelo:
Disregard.
The real issue is in our app, fixed here:
`https://github.com/IBM/watson-online-store/pull/135`

ehdsouza:
@scottdangelo , there was a problem in the swagger, check https://github.ibm.com/Watson/developer-cloud--api-explorer/pull/366#event-22362980

I have regenerated conversation and assistant:
PR: https://github.com/watson-developer-cloud/python-sdk/pull/413
PR: https://github.com/watson-developer-cloud/python-sdk/pull/414

scottdangelo:
Thanks. Yes, that PR #413 looks like it would fix our issue.
Thanks for getting back to me on this.

PR #411, opened 2018-03-16
Use nested dict to access `conversation_id`.

scottdangelo:
The checks for `conversation_id` in:
`watson_developer_cloud/conversation_v1.py", line 1716, in _from_dict"`
Fails because it is a nested dictionary within a dictionary.
We must access the `conversation_id` field via:
`['context']['conversation_id']`

Closes: #410

### DIFF
--- a/watson_developer_cloud/conversation_v1.py
+++ b/watson_developer_cloud/conversation_v1.py
@@ -1708,16 +1708,16 @@ class Context(object):
         """Initialize a Context object from a json dictionary."""
         args = {}
         xtra = _dict.copy()
-        if 'conversation_id' in _dict:
-            args['conversation_id'] = _dict['conversation_id']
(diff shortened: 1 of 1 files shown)


ehdsouza:
@scottdangelo is this valid?

scottdangelo:
No, I think we need to fix it in our app.
But....the previous version of the SDK allowed us to send the entire message, with ['context']['conversation_id'] nested within and no failure.
Now, the new SDK checks for 'conversation_id', but requires the 'context' to be stripped out of the message.
It would be nice if we could still just sent the entire JSON blob, like we used to.
I'm guessing that a POST to the Service endpoint can operate on the full message? Without sending context as a parameter to the SDK message() call?
Anyway, I'll close this for now. I'll get back to this Monday if we discover anything new.
Thanks.

Issue #402, opened 2018-03-12
Hang when running the WebSocket speech-to-text example


Omegastick:
#### Expected behavior
Running the speech to text example, the message recognized from the audio should be retrieved from IBM Watson twice (once for the HTTP API and once for the WebSocket API)

#### Actual behavior
The HTTP API runs normally, but the WebSocket part hangs indefinitely after `speech-to-text.recognize_with_websocket()` is called.

Adding an `inactivity_timeout` parameter does nothing.

#### Steps to reproduce the problem
Install the Watson Python SDK and it's dependencies. Run the example speech to text script.
I assume it was tested working when it was committed, so I assume there is something else at work here.
I'm behind a company proxy which has blocked WebSocket connections in the past sometimes, but the JavaScript browser demo (https://speech-to-text-demo.ng.bluemix.net/) works fine.

#### Code snippet (Note: Do not paste your credentials)
samples/speech_to_text_v1.py verbatim except for the audio file path is replaced with one on my machine.

#### python sdk version
1.1.1

#### python version
2.7.14

ehdsouza:
Hi @Omegastick,

Did you get the `Connection was successful` response? That is the initial handshake that happens with the service.

I am suspecting it to be a proxy issue as the python sdk uses autobahn and is independent of the demo working.

ehdsouza:
Before running the example, could you run this, to enable http logging:
 
```py
try:
    import http.client as http_client
 except ImportError:
    # Python 2
    import httplib as http_client
 http_client.HTTPConnection.debuglevel = 1
```

Omegastick:
Running with the original audio file and with HTTP logging enabled:
```
send: 'CONNECT stream.watsonplatform.net:443 HTTP/1.0\r\n'
send: '\r\n'
send: 'GET /speech-to-text/api/v1/models HTTP/1.1\r\nHost: stream.watsonplatform.net\r\nConnection: keep-alive\r\nAccept-Encoding: gzip, deflate\r\naccept: application/json\r\nuser-agent: watson-apis-python-sdk-1.1.1 Linux 4.13.0-36-generic 2.7.14\r\nAuthorization: Basic [Authorization]==\r\n\r\n'
reply: 'HTTP/1.1 200 OK\r\n'
header: X-Backside-Transport: OK OK
header: X-Frame-Options: DENY
header: X-Content-Type-Options: nosniff
header: Content-Security-Policy: default-src
[...]

PR #421, opened 2018-03-27
Added proxy setting for speech_to_text WebSockets

Omegastick:


germanattanasio:
Hey @Omegastick you need to sign the commit and probably add a test before we can merge this. Check your git configuration

Omegastick:
I think I fixed the username problem.

I'm having trouble adding a test because there's no mock library for your WebSocket implementation (as far as I can see). I'd have to add another dependency. Any suggestions?

By the way, as mentioned in #402, this doesn't actually work for me. The bug I was getting before is still there.

germanattanasio:
We currently have a way to setup the http configuration parameters. See https://github.com/watson-developer-cloud/python-sdk#configuring-the-http-client-supported-from-v110

I would like to use the parameters there instead of asking the user for the proxy host and port again.

Omegastick:
Would you like me to add a `set_websocket_config` method?

germanattanasio:
I think we can add a `service.set_websocket_config({'proxy': { ... } })`

Omegastick:
Sorry for the delay, work has been busy lately. I'll get around to this this week I think.

ehdsouza:
Closing this as we would be looking into a new web socket library

PR #423, opened 2018-03-27
Classify multiple lines of text

gfoster1:
As a user I want to classify multiple lines of text in one api call.


ehdsouza:
Hi @gfoster1,

Really appreciate your effort and time in this PR. Currently, we automatically generate the methods using swagger which generate input/ output as models. I created a PR: https://github.com/watson-developer-cloud/python-sdk/pull/424 and used your unit test example.

Thank you once again!

PR #427, opened 2018-03-29
feat(core_ml): Add get_core_ml_model in VR

ehdsouza:


germanattanasio:
@ehdsouza you need to find why you can't deploy the library to pip. Make sure you update the CHANGELOG too

PR #434, opened 2018-04-05
:fix (VR): bug fixes in VR

ehdsouza:
1) Added a comment for example of stt using websockets where only initial recording is allowed
2) VR changes to take the new way of taking params
3) Commented core-ml example as it downloads a file


germanattanasio:
@ehdsouza make sure the sdk generation takes this into account and generates the `isInstance()`

PR #438, opened 2018-04-10
Uniform request handling

justinas-vd:
This patch returns the response object. The motivation for this patch is that I use the response object from the self.request.

I use it because I perform asynchronous requests (by overriding the request method in a manner like this):

```Python
def make_async_service(watson_service):
    "Override the request method, so that it acts nicer within the async environment."
    watson_service.request = types.MethodType(async_request, watson_service)
    watson_service._tls = threading.local()
    return watson_service
```

Everything works except for the `delete_classifier` function because for some arbitrary reason it returns `None` instead  of the response. This is an API bug.

### DIFF
--- a/watson_developer_cloud/natural_language_classifier_v1.py
+++ b/watson_developer_cloud/natural_language_classifier_v1.py
@@ -218,9 +218,9 @@ class NaturalLanguageClassifierV1(WatsonService):
             headers.update(kwargs.get('headers'))
         url = '/v1/classifiers/{0}'.format(
             *self._encode_path_vars(classifier_id))
-        self.request(
+        response = self.request(
             method='DELETE', url=url, headers=headers, accept_json=True)
-        return None
+        return response
 
     def get_classifier(self, classifier_id, **kwargs):
         """


germanattanasio:
@justinas-vd Thanks for the PR! It looks good but I don't think we want to expose the request response.
I'm interested in your use case and the fact that you need to support asynchronous requests. Can you please elaborate on that?

BTW: The support for asynchronous requests in the Java SDK started from a user request like this

justinas-vd:
Not exposing the request response simply forces me to duplicate the code in the `delete_classifier` fn. Returning `None` in that context does not have any objective benefit – neither it makes the API more secure, nor is it cheaper. 

As for the asynchronous requests, it's a bit more complex in Python than in Java, because Java has a clearly defined standard async API. Python is a bit more eclectic, which is both a blessing and a curse.
What I did, though, I simply wrote `async_request` function. A working example implemented using aiohttp would look like this

```Python
async def perform_async_request(watson_service, request_func):
    "Use cached session in `watson_service` to perform a request."
    # Use cached sessions whenever possible
    # pylint: disable=bare-except
    try:
        session = watson_service._tls.client_session[-1].get_session()
    except:
        async with aiohttp.ClientSession(cookies=watson_service.jar) as session:
            return await request_func(session)
    else:
        return await request_func(session)

# pylint: disable=too-many-arguments
async def async_request(self, method, url, accept_json=False, headers=None,
                        params=None, json=None, data=None, files=None, **kwargs):
    """
    This is an async implementation of watson_developer_cloud_service.request method,
    so that it better meshes with the general aiohttp usage scenario. We don't want
    our requests to Watson to be blocking.
    """

    full_url = self.url + url

    input_headers = _remove_null_values(headers) if headers else {}

    headers = CaseInsensitiveDict(self.user_agent_header)
    if self.default_headers is not None:
        headers.update(self.default_headers)
    if accept_json:
        headers['accept'] = 'application/json'
    headers.update(input_headers)

    # Remove keys with None values
    params = _remove_null_values(params)
    params = _cleanup_param_values(params)
    json = _remove_null_values(json)
    data = _remove_null_values(data)
    files = _remove_null_values(files)

    if sys.version_info >= (3, 0) and isinstance(data, str):
        data = data.encode('utf-8')

    # Support versions of requests older than 2.4.2 without the json input
    if not data and json is not None:
        data = json_import.dumps(json)
        headers.update({'content-type': 'application/json'})

    # Handle multipart form data
    elif files:
        if isinstance(data, str):
            raise ValueError("Data must not be a string.")

        form = aiohttp.FormData()

        # Data can only be 1-level dictionary there.
        # It cannot contain nested dictionaries.
        form.add_fields(MultiDict(data or {}))

        files = to_key_val_list(files or {})

        # pylint: disable=C0103
        for (k, v) in files:
            # Support for explicit filename
            ft = None
            fh = None
            if isinstance(v, (tuple, list)):
                if len(v) == 2:
                    fn, fp = v
                elif len(v) == 3:
                    fn, fp, ft = v
                else:
                    fn, fp, ft, fh = v
            else:
                fn = guess_filename(v) or k
                fp = v

            form.add_field(k, value=fp, content_type=ft, filename=fn, content_transfer_encoding=fh)

        data = form


    auth = None
    if self.username and self.password:
        auth = aiohttp.BasicAuth(self.username, self.password, encoding='utf-8')
    if self.api_key is not None:
        if params is None:
            params = {}
        if full_url.startswith(
                'https://gateway-a.watsonplatform.net/calls'):
            params['apikey'] = self.api_key
        else:
            params['api_key'] = self.api_key

    async def _session_request(session):
        async with session.request(method=method, url=full_url,
                                   auth=auth,
                                   headers=headers,
                                   params=params, data=data,
                                   **kwargs) as response:

            if 200 <= response.status <= 299:
                if accept_json:
                    response_json = await response.json(encoding='utf-8')
                    if 'status' in response_json and response_json['status'] \
                            == 'ERROR':
                        status_code = 400
                        error_message = 'Unknown error'

                        if 'statusInfo' in response_json:
                            error_message = response_json['statusInfo']
                        if error_message == 'invalid-api-key':
                            status_code = 401
                        raise WatsonApiException(status_code, error_message)
                    return response_json
                return await convert_to_requests_response(response)
            else:
                # Intentionally convert response type here, so that it is easier to
                # plug-in the original error handling code here.
                response = await convert_to_requests_response(response)
                raisable_exception_type = WatsonApiException
                if response.status_code == 401:
                    error_message = 'Unauthorized: Access is denied due to ' \
                                    'invalid credentials'
                else:
                    if response.status_code == 404:
                        raisable_exception_type = WatsonServiceNotFoundException
                    error_message = self._get_error_message(response)
                error_info = self._get_error_info(response)
                raise raisable_exception_type(response.status_code, error_message,
                                              error_info)

    return await perform_async_request(self, _session_request)

def make_async_service(watson_service):
    "Override the request method, so that it acts nicer within the async environment."
    watson_service.request = types.MethodType(async_request, watson_service)
    watson_service._tls = threading.local()
    return watson_service
```

This function works almost the same as self.request, except it is asyncronous. In order to get the result you need to `await` the request. For the most part this kind of injection works wonderfully for both NLC and conversation (I didn't try other services), but you need to transparently return the request objects.

I will note, though, that aiohttp is just one of the many async http libraries in Python, so I don't know how much you would like to tie yourself with a concrete http backend. On the other hand, even though my approach is kinda hackish, but it works and in my case I need just-a-wee-bit-help to make me happy all the way.

justinas-vd:
Sure. BTW, just yesterday I also used ToneAnalyzer asynchronously. Worked like a charm, too.

From my point of view, I can't wait until the next major release, so at least don't hinder what works, okay? Please either merge those two lines of code or put them in yourselves.

germanattanasio:
@justinas-vd besides we were returning `None` for the `delete` operation, I don't think we want to return the `requests` response because that will expose our internal implementation.

I was thinking of exposing this functionality without having to expose that we use `requests`

justinas-vd:
This will not expose `requests` per se, since `accept_json=True`. Normally, this would return a JSON response from self.request.

mkistler:
I'm not sure what the problem is here.  It is perfectly valid for an API to return no response -- there is a special status code just for that -- 204 No Content.

Now it is true that in this case the service is returning a 200 and an empty response, but since this is effectively "No Content", the SDK returns "None" to signify that.

If your application requires "some response" -- even if it is empty -- you can accomplish this yourself:
```
>>> response = None
>>> whatIneed = response or {}
>>> whatIneed
{}
>>> 
```

ehdsouza:
I think we would need to support uniform response handling for even the `DELETE` api calls especially if we want to support returning response headers for all the methods.

PR #451, opened 2018-04-17
feat(readme): Update readme with gdpr support

ehdsouza:


watson_developer_cloud/watson_service.py
```diff
@@ -145,15 +145,21 @@ class DetailedResponse(object):
...
+
+    def get_headers(self):
+        return self.headers
+
     def _to_dict(self):
         _dict = {}
-        if hasattr(self, 'response') and self.response is not None:
-            _dict['response'] = self.response
+        if hasattr(self, 'result') and self.result is not None:
+            _dict['result'] = self.result
```
germanattanasio: You need to check if the result can be stringified or not... Maybe it's a stream in which case it won't be stringified.

PR #462, opened 2018-05-14
Uniform request handling

justinas-vd:
Allow self.request to return an async coroutine, which then can be awaited.

Nota bene: none of these changes expose requests interface, since all of the update methods actually return a JSON dict. Moreover, it is unclear why this dict should not be returned in the sync case.

### DIFF
--- a/test/unit/test_assistant_v1.py
+++ b/test/unit/test_assistant_v1.py
@@ -105,7 +105,6 @@ def test_delete_counterexample():
         workspace_id='boguswid', text='I want financial advice today')
     assert len(responses.calls) == 1
     assert responses.calls[0].request.url.startswith(url)
-    assert counterexample is None
 
 
 @responses.activate
@@ -241,7 +240,7 @@ def test_create_entity():
 def test_delete_entity():
     endpoint = '/v1/workspaces/{0}/entities/{1}'.format('boguswid', 'pizza_toppings')
     url = '{0}{1}'.format(base_url, endpoint)
-    response = ""
+    response = {}
     responses.add(
         responses.DELETE,
         url,
@@ -253,7 +252,6 @@ def test_delete_entity():
     entity = service.delete_entity(workspace_id='boguswid', entity='pizza_toppings')
     assert len(responses.calls) == 1
     assert responses.calls[0].request.url.startswith(url)
-    assert entity is None
 
 
 @responses.activate
@@ -414,7 +412,6 @@ def test_delete_example():
         text='Gimme a pizza with pepperoni')
     assert len(responses.calls) == 1
     assert responses.calls[0].request.url.startswith(url)
-    assert example is None
 
 
 @responses.activate
@@ -567,7 +564,6 @@ def test_delete_intent():
         workspace_id='boguswid', intent='pizza_order')
     assert len(responses.calls) == 1
     assert responses.calls[0].request.url.startswith(url)
-    assert intent is None
 
 
 @responses.activate
@@ -998,7 +994,7 @@ def test_delete_synonym():
     endpoint = '/v1/workspaces/{0}/entities/{1}/values/{2}/synonyms/{3}'.format(
         'boguswid', 'aeiou', 'vowel', 'a')
     url = '{0}{1}'.format(base_url, endpoint)
-    response = ""
+    response = {}
     responses.add(
         responses.DELETE,
         url,
@@ -1011,7 +1007,6 @@ def test_delete_synonym():
         workspace_id='boguswid', entity='aeiou', value='vowel', synonym='a')
     assert len(responses.calls) == 1
     assert responses.calls[0].request.url.startswith(url)
-    assert synonym is None
 
 
 @responses.activate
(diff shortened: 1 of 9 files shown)


germanattanasio:
@justinas-vd we are not going to return the response object since it will expose our underlying library. We have introduced a `DetailedResponse` that you can see as an interface to access the response without exposing `requests` maybe we can update that class to expose the async information you need.

justinas-vd:
First, **None of the changes expose the underlying library.**

This is precisely because the `self.request` method does not return a request object (to which you object, pardon the pun) when `accept_json=True ` is specified. In that case it simply returns a parsed JSON dictionary.

So if you wanted to say that something is "exposed", then the only thing is exposed is a JSON dictionary. Which, by the way, should be exposed. Why on Earth I cannot inspect a JSON response from a Watson service -- I have no idea. I am quite sure, though, that the devs should not prevent me from doing that.


Second, **DetailedResponse does not help if you don't return it.**

When you explicitly return `None` from your API functions, I can have neither detailed responses, nor async coroutine handles upon which I could `await`.

To put it bluntly: fix that brain damage with "return None". It was simply an ill-conceived idea.

germanattanasio:
@justinas-vd The `DetailedResponse` is not returned unless you specify that you want it with `set_detailed_response(True)`

I just went to see what `response = self.request()` was doing and you are right about the dictionary. I thought it was the response from `request()` but that's because we are calling the method `request()` 🤦‍♂️ 

I think we can do this because currently the method ALWAYS returns None, so I don't expect people are writing code based on the response.

Please give me a few days so that I can ask someone else before merging this. We will also have to update a template we use internally to generate the code so that we don't override this manual change.

justinas-vd:
Thank you. I will also update unit tests so that travis does not complain.

mkistler:
There is definitely a problem here, though it’s a bit more complicated than what you suggest.

1.  It does appear that we are not returning a `DetailedReponse` object back on `DELETE`s when detailed responses have been requested, and that is broken.  We need to fix that.
2. When `detailedResponse` is _not_ set, we should continue returning `None`.  We probably have to do that in the method itself, with logic something like
    ```
    response = self.request( ...
    return response if self.detailed_response else None
    ```

3. The `:rtype:`  in the doc comments all need to be fixed.  It should either be
`:rtype: Union[None, DetailedResponse]`
or
`:rtype: Union[dict, DetailedResponse]`

And of course, we need to fix the generator to do this the right way for the next time we generate the Python SDK.

justinas-vd:
Nope. No "None", thank you. This "None" gives only problems and solves absolutely nothing.

Even your suggestion to return None when `DetailedResponse` is not requested is kinda kludgy. Why would one want to write `return response if self.detailed_response else None` in literally dozens of places?

Also, why as a library user I cannot get the JSON from Watson services? I want to see if my request to delete a NLC instance is successful, I want to be able to verify that I deleted a synonym or whatnot. This artificial `return of None` unnecessarily (and for ideological reasons) limits the Python API. It is my discretion -- as a user's -- to ignore what Watson services say back when I delete or put something. Not yours.

jsstylos:
I support moving from `None` to `{}` for the sake of consistency and evolvability. (The initial design did just this.)

Shouldn't changing it back be considered a breaking change, given that users may have their own tests / etc. that check for `None` with the current version?

justinas-vd:
Most likely it won't be a breaking change. A breaking change was introduced with returning `None`. I was almost apoplectic when I got this "new and improved" version. It broke my own project at IBM, and currently I have to duplicate the deletion code.

jsstylos:
As your experience shows, changing the response type can definitely break code. The question here is whether changing it back will break more people who have already written new code, or break more people who haven't yet adapted to the new code.

justinas-vd:
I agree with your concern but I will note this:

1) it's not possible to recover the needed value when you return None;

2) it's quite possible to comment out the assert if you expected None and now you get something more.

So in that regard the breakages are not all equal. The first kind if much much worse than the second.

ehdsouza:
HI @justinas-vd, thanks a lot for your PR. We have an issue created internally to support uniform request handling along with generator changes. 
Also, this would be a breaking change as the `DELETE` response is now changed from `None` to `{}`

mkistler:
@justinas-vd, when we fix the `detailedResponse` logic, the SDK will return a `DetailedResponse` from _every_ service call when `detailedResponse` is set.  Will this be sufficient for your needs?

justinas-vd:
@mkistler   If I understand the vision that you have laid out in README correctly, then the answer would be a "yes".

My requirement is simple: every service call must return a "response" from `self.request` (be it a `{}`, a `DetailedResponse` -- I don't really care). Essentially, I need to use that response to piggyback my async coroutines. (Which I absolutely cannot do when you plainly return `None` and thus discard the result from `self.request`.)

BTW, this PR adresses exactly that and the patched functions would return `DetailedResponse`, too, when you update `self.request`.

ehdsouza:
@justinas-vd thanks for taking the time for making the change.
We now have it in the latest `v2.0.0` release. Check the https://github.com/watson-developer-cloud/python-sdk/wiki/Migration for details

justinas-vd:
@ehdsouza Excellent news. Thank you.

PR #499, opened 2018-06-26
add Language Translator V3 examples

g-may:


ehdsouza:
@g-may the examples are part of integration tests. So, the travis env file would need to be updated as well.

PR #503, opened 2018-06-29
Language translator V2 deprecation notice

ehdsouza:


ehdsouza:
@SirSpidey thanks for the suggestion :-), updated the readme

PR #514, opened 2018-07-11
Add access token Auth for web socket

ehdsouza:


watson_developer_cloud/speech_to_text_v1.py
```diff
@@ -516,9 +517,13 @@ def recognize_with_websocket(self,
         if 'headers' in kwargs:
             headers.update(kwargs.get('headers'))
 
-        authstring = "{0}:{1}".format(self.username, self.password)
-        base64_authorization = base64.b64encode(authstring.encode('utf-8')).decode('utf-8')
-        headers['Authorization'] = 'Basic {0}'.format(base64_authorization)
+        if self.token_manager:
```
germanattanasio: What happens if the user creates an instance of STT and waits more than 1 hour before calling a method? will that trigger a token refresh?

ehdsouza: `get_token()` would be doing the logic of refreshing the token

PR #518, opened 2018-07-16
refactor(Add adapters for hand edits):

ehdsouza:
* Add speech_to_text adapter that contains the web socket recognize method
* Add visual_recognition adapter that contains the create_classifier and update_classifier


watson_developer_cloud/__init__.py
```diff
@@ -27,9 +27,9 @@
 from .natural_language_understanding_v1 import NaturalLanguageUnderstandingV1
 from .personality_insights_v2 import PersonalityInsightsV2
 from .personality_insights_v3 import PersonalityInsightsV3
-from .speech_to_text_v1 import SpeechToTextV1
+from .speech_to_text_v1_adapter import SpeechToTextV1Adapter as SpeechToTextV1
```
germanattanasio: Maybe it will be good to move all the adapter to the bottom of this file and add some lines describing the reasons

ehdsouza:
@germanattanasio the generator would add `create_classifier` and `update_classifier` in the generated file which is removed in my generation script using this simple `sed` command:
`sed  -i '' '/create_classifier/,/return response/d' <vr_generated_file>`

The Speech to text wouldnt need any deletion.

Issue #539, opened 2018-08-17
Error in python using speech_to_text websocket when using a recognize_callback, interim_results and speaker_labels


Peamon:
#### Expected behavior
Getting json result

#### Actual behavior
2 Hunhandled Errors

```
Unhandled Error
Traceback (most recent call last):
  File "/home/clarte/miniconda3/envs/track/lib/python3.5/site-packages/twisted/python/log.py", line 103, in callWithLogger
    return callWithContext({"system": lp}, func, *args, **kw)
  File "/home/clarte/miniconda3/envs/track/lib/python3.5/site-packages/twisted/python/log.py", line 86, in callWithContext
    return context.call({ILogContext: newCtx}, func, *args, **kw)
  File "/home/clarte/miniconda3/envs/track/lib/python3.5/site-packages/twisted/python/context.py", line 122, in callWithContext
    return self.currentContext().callWithContext(ctx, func, *args, **kw)
  File "/home/clarte/miniconda3/envs/track/lib/python3.5/site-packages/twisted/python/context.py", line 85, in callWithContext
    return func(*args,**kw)
--- <exception caught here> ---
  File "/home/clarte/miniconda3/envs/track/lib/python3.5/site-packages/twisted/internet/posixbase.py", line 614, in _doReadOrWrite
    why = selectable.doRead()
  File "/home/clarte/miniconda3/envs/track/lib/python3.5/site-packages/twisted/internet/tcp.py", line 243, in doRead
    return self._dataReceived(data)
  File "/home/clarte/miniconda3/envs/track/lib/python3.5/site-packages/twisted/internet/tcp.py", line 249, in _dataReceived
    rval = self.protocol.dataReceived(data)
  File "/home/clarte/miniconda3/envs/track/lib/python3.5/site-packages/twisted/protocols/tls.py", line 330, in dataReceived
    self._flushReceiveBIO()
  File "/home/clarte/miniconda3/envs/track/lib/python3.5/site-packages/twisted/protocols/tls.py", line 295, in _flushReceiveBIO
    ProtocolWrapper.dataReceived(self, bytes)
  File "/home/clarte/miniconda3/envs/track/lib/python3.5/site-packages/twisted/protocols/policies.py", line 120, in dataReceived
    self.wrappedProtocol.dataReceived(data)
  File "/home/clarte/miniconda3/envs/track/lib/python3.5/site-packages/autobahn/twisted/websocket.py", line 131, in dataReceived
[...]

PR #540, opened 2018-08-17
new(WS): web socket-client library for STT weboscket

ehdsouza:
This PR replaces the `autobahn`library with `websocket-client`.

Added a new class `AudioSource` which would represent an audio file or a buffer in the form of a queue for a microphone streaming.

Added integration tests for an audio file.

Fixes: https://github.com/watson-developer-cloud/python-sdk/issues/539


watson_developer_cloud/websocket/recognize_listener.py
```diff
@@ -0,0 +1,199 @@
...
+    def on_data(self, ws, message, message_type, fin):
+        """
+        Callback executed when message is received from the server.
+
+        :param ws: Websocket client
+        :param message: utf-8 string which we get from the server.
+        :param message_type: Message type which is either ABNF.OPCODE_TEXT or ABNF.OPCODE_BINARY
+        :param fin: continue flag. If 0, the data continues.
+        """
+        json_object = json.loads(message)
```
germanattanasio: Sometimes the data can't be parsed. You need to take that into account. Maybe a try catch

PR #580, opened 2018-10-16
fix(docs): incorrect parameter types in STT

PaperBag42:
In the documentation for `SpeechToTextV1`'s `recognize`, `create_job` and `add_audio`, the audio parameter was labeled as `str` where it should be a file.


germanattanasio:
Thanks @PaperBag42

PR #598, opened 2018-11-20
fix(all services): Don't try to deserialize a body on HEAD requests

ricellis:
### Summary

The `watson_developer_cloud.watson_service.WatsonService#request` method throws an exception with HTTP `HEAD` requests as it tries to deserialize a `None` body:

```
  File "watson_developer_cloud/watson_service.py", line 469, in request
    response_json = response.json()
  File "requests/models.py", line 897, in json
    return complexjson.loads(self.text, **kwargs)
  File "json/__init__.py", line 348, in loads
    return _default_decoder.decode(s)
  File "json/decoder.py", line 337, in decode
    obj, end = self.raw_decode(s, idx=_w(s, 0).end())
  File "json/decoder.py", line 355, in raw_decode
    raise JSONDecodeError("Expecting value", s, err.value) from None
json.decoder.JSONDecodeError: Expecting value: line 1 column 1 (char 0)
```

The `WatsonService#request` method already handles a no body response case for a `204` status code, so this PR adds an additional condition to use that path for `HEAD` HTTP method requests.

Added a new test `test.unit.test_watson_service.test_http_head` that validates an exception is not thrown and asserts that the response headers can be retrieved from the `DetailedResponse` to a `HEAD` request.


germanattanasio:
Thanks  @ricellis

PR #621, opened 2019-01-07
chore(all services): Check for incorrect url and credentials

ehdsouza:
Users have had issues with authenticating services if they mistakenly add { or " into different fields as described in documentation. We should detect these in the first character of the username, password, apikey and service url and throw an error.


mediumTaj:
@ehdsouza These tests keep failing. I separated the assistant v2 readme change to another PR. Can you take a look at this when you get back?

PR #624, opened 2019-01-10
Apply manual changes for DialogNode

mediumTaj:
### Summary
This pull request moves the `disabled` field from the `CreateDialogNode` model to the `DialogNode` model as described in the API definition.


mediumTaj:
This is failing because of a service issue where `GET /v1/recognitions` times out. The service team is looking into it.

PR #634, opened 2019-01-30
Allow newer versions of websocket-client

jhoelzl:
In order to prevent conflicts with other python modules


ehdsouza:
@jhoelzl thanks for the PR 👍
We wouldn't want to get the latest version as the `websocket-client` library is still in beta and not all versions look stable.

Issue #644, opened 2019-03-12
SpeechToTextV1: Error when authenticating with ibm-credentials.env


marxemat:
**Overview**
SpeechToTextV1 Service with ibm-credentials.env file

**Expected behavior**
SpeechToTextV1 (with apikey and url) should be initiated and recognize should work  

**Actual behavior**
`watson_developer_cloud.watson_service.WatsonApiException: Error: Provided API key could not be found, Code: 400`
When copying the api key and url directly to `SpeechToTextV1(url="", iam_apikey="")` it works fine without failures.

**How to reproduce**
Environment variable `IBM_CREDENTIALS_FILE` is set and accessed, otherwise it would be a different error. I have an apikey with upper and lowercase characters and numbers.

```
service = SpeechToTextV1()

with open(join(dirname(__file__), './.', 'C:\\Users\\...), 'rb') as audio_file:
    speech_recognition_results = service.recognize(
        audio=audio_file,
        content_type='audio/wav',
        word_alternatives_threshold=0.9,
        word_confidence=True,
        timestamps=True,
        profanity_filter=False,
        smart_formatting=True,
        speaker_labels=True
    ).get_result()
```

I could determine that the apikey is lowered before submitting in the request.

In watson_service.py I found:
```
def load_from_credential_file(self, service_name, separator='='):
        […]
        if credential_file_path is not None:
            with open(credential_file_path, 'r') as fp:
                for line in fp:
                    key_val = line.strip().split(separator)
                    if len(key_val) == 2:
                        self._set_credential_based_on_type(service_name, key_val[0].lower(), key_val[1].lower())
```

When changing the last line to

` self._set_credential_based_on_type(service_name, key_val[0].lower(), key_val[1])`

the program now runs without failure. 

**SDK Version**
2.8.0

**Additional information:**
- OS: Windows x64
- Python: 3.7

germanattanasio:
OMG 🤦‍♂️ I'm sorry @marxemat I don't know how we ended up transforming an apikey to lowercase. Since you found the problem and solution, do you mind writing a PR?

ehdsouza:
🙍 😞!! Thanks for debugging this and getting the PR out 😀

ehdsouza:
fixed with PR: https://github.com/watson-developer-cloud/python-sdk/pull/645

PR #645, opened 2019-03-13
fix(service): transform apikey to lowercase

marxemat:
see https://github.com/watson-developer-cloud/python-sdk/issues/644


ehdsouza:
@marxemat thanks for catching this bug 🥇 

The PR looks good, could you amend your commit message to be something like: `fix(service): transforming apikey to lowercase` as we use semantic release and we need to follow angular commit messages https://github.com/angular/angular/blob/master/CONTRIBUTING.md#-commit-message-guidelines for a release to take place.

PR #646, opened 2019-03-15
chore(common): Create a common file

ehdsouza:
Create a common file to allow SDK to provide a set of default headers with each REST API request

### DIFF
--- a/watson_developer_cloud/common.py
+++ b/watson_developer_cloud/common.py
@@ -1,0 +1,41 @@
+# coding: utf-8
+
+# Copyright 2019 IBM All Rights Reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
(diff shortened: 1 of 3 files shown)


watson_developer_cloud/common.py
```diff
@@ -0,0 +1,40 @@
...
+USER_AGENT_HEADER = 'User-Agent'
+SDK_NAME = 'watson-apis-python-sdk'
+
+def get_system_info():
+    return '{0} {1} {2}'.format(platform.system(), # OS
+                                platform.release(), # OS version
+                                platform.python_version()) # Python version
+
+def get_user_agent():
+    return '{0}-{1} {2}'.format(SDK_NAME, __version__, get_system_info())
```
padamstx: In the area of performance... it would probably be better if we only computed this string once the first time the function was called and then cached the value for each subsequent call. That way we're not computing the string EVERY time a REST API is invoked.

ehdsouza: Good point 👍 I updated it to be in a global variable.

PR #647, opened 2019-03-15
fix(example): Update ta and conversation example

ehdsouza:
Updated conversation and ta example to use assistant instead.

Fixes I had to do:
* Update TA version to `version='2016-05-19'` as the 3 categories are deprecated, see notice here:
```
:param list[str] tones: **`2017-09-21`:** Deprecated. The service continues to
        accept the parameter for backward-compatibility, but the parameter no longer
        affects the response.
        **`2016-05-19`:** A comma-separated list of tones for which the service is to
        return its analysis of the input; the indicated tones apply both to the full
        document and to individual sentences of the document. You can specify one or more
        of the valid values. Omit the parameter to request results for all three tones.
        :param str content_language: The language of the input text for the request:
        English or French. Regional variants are treated as their parent language; for
        example, `en-US` is interpreted as `en`. The input content must match the
        specified language. Do not submit content that contains both languages. You can
        use different languages for **Content-Language** and **Accept-Language**.
```

* In `tone_detection.py` the categories returned were:(note there is no writing_tone)
```
EMOTION_TONE_LABEL = 'emotion_tone'
WRITING_TONE_LABEL = 'language_tone'
SOCIAL_TONE_LABEL = 'social_tone'
```


examples/assistant_tone_analyzer_integration/tone_assistant_integration.v1.py
```diff
@@ -13,17 +13,17 @@
...
 
-# replace with your own conversation credentials or put them in a .env file
-conversation = ConversationV1(
-    username=os.environ.get('CONVERSATION_USERNAME') or 'YOUR SERVICE NAME',
-    password=os.environ.get('CONVERSATION_PASSWORD') or 'YOUR PASSWORD',
-    version='2016-09-20')
+# replace with your own assistant credentials or put them in a .env file
+assistant = AssistantV1(
+    username=os.environ.get('ASSISTANT_USERNAME') or 'YOUR SERVICE USERNAME',
+    password=os.environ.get('ASSISTAN_PASSWORD') or 'YOUR SERVICE PASSWORD',
```
mediumTaj: Missing a `T` here in `ASSISTAN_PASSWORD`

aprilwebster:
Sorry to come to the party late.  LGTM too.  The only minor thing I was looking into is the `WRITING_TONE_LABEL = 'language_tone'`.  Asked someone on my team closer to the service whether `WRITING` should be switched to `LANGUAGE` at some point.

PR #654, opened 2019-03-29
Display moving to ibm-watson

ehdsouza:
This PR displays a warning to a user about the renaming of package

### DIFF
--- a/README.md
+++ b/README.md
@@ -36,6 +36,7 @@ Python client library to quickly get started with the various [Watson APIs][wdc]
 * You need an [IBM Cloud][ibm-cloud-onboarding] account.
 
 ## Installation
+Note: We are renaming `watson-developer-cloud` to `ibm-watson`
 
 To install, use `pip` or `easy_install`:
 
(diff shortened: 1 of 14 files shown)


README.md
```diff
@@ -36,6 +36,7 @@ Python client library to quickly get started with the various [Watson APIs][wdc]
 * You need an [IBM Cloud][ibm-cloud-onboarding] account.
 
 ## Installation
+Note: We are renaming `watson-developer-cloud` to `ibm-watson`
```
mediumTaj: Should this say `renaming`? Maybe we should say something like we are moving to `ibm-watson` and let them know that all versions prior to `v3.0.0` can still be found in `watson-developer-cloud`?

watson_developer_cloud/speech_to_text_v1.py
```diff
@@ -44,7 +44,6 @@
 # Service
 ##############################################################################
 
-
 class SpeechToTextV1(WatsonService):
```
mediumTaj: Should there be deprecation notice here?

watson_developer_cloud/text_to_speech_v1.py
```diff
@@ -40,7 +40,6 @@
 # Service
 ##############################################################################
 
-
 class TextToSpeechV1(WatsonService):
```
mediumTaj: Should there be deprecation notice here?

README.md
ehdsouza: ok sounds good

PR #661, opened 2019-04-03
Drop support for Python 3.4 because it is now EOL

cclauss:
Remove support for Python 3.4 because that release has reached its end of life.
* https://devguide.python.org/devcycle/#end-of-life-branches


cclauss:
Codacy seems to be improperly configured so perhaps it should be disabled.

ehdsouza:
@cclauss thanks for the PR 🥇

PR #680, opened 2019-07-29
Python SDK Dockerfile

kingsodigie:
Added dockerfile to python-sdk which creates an image with the sdk installed

### DIFF
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -3,7 +3,8 @@ FROM python:3-slim
 
 #Install Git
 RUN apt-get update && \
-    apt-get install -y --no-install-recommends git && \
+    apt list -a git &&  \
+    apt-get install -y --no-install-recommends git=1:2.20.1-2 && \
     apt-get clean && \
     rm -rf /var/lib/apt/lists/*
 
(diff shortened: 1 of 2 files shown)


docker/Dockerfile
```diff
@@ -0,0 +1,21 @@
...
+
+#Clone SDK repo to app directory
+RUN git clone https://github.com/watson-developer-cloud/python-sdk.git
+
+#Set working directory to python-sdk
+WORKDIR /python-sdk
+
+#Pip install dependencies for SDK
+RUN pip install --editable .
+RUN pip install -r requirements-dev.txt
```
germanattanasio: This will not install the SDK as a user but as a developer

docker/Dockerfile
```diff
@@ -0,0 +1,5 @@
+#Import Python Base Image(change version as needed)
+FROM python:3-slim
+
+#Install the python SDK
+RUN pip install --upgrade ibm-watson==3.2.0
```
ehdsouza: Keep this to `ibm-watson>=3.2.0`. 

Ideally we would want bumpversion to update it as well, but its ok for now.

PR #694, opened 2019-10-01
Visual recognition v4

ehdsouza:
* Adds support for visual recognition v4
* Also generated using the new format where constructors no longer have the url param
* Updated examples and tests


mediumTaj:
Erika can you delete visrec models and regenerate with api def `09a54edf1fa2f715752a8d9f6f7d942c3cb4d8b4`

ibm_watson/visual_recognition_v4.py
```diff
@@ -0,0 +1,2480 @@
...
+        sdk_headers = get_sdk_headers('watson_vision_combined', 'V4', 'analyze')
+        headers.update(sdk_headers)
+
+        params = {'version': self.version}
+
+        form_data = []
+        for item in collection_ids:
+            form_data.append(('collection_ids', (None, item, 'text/plain')))
+        for item in features:
+            form_data.append(('features', (None, item, 'text/plain')))
```
mediumTaj: Do we need to convert collection_ids and features into csv here? I think they are lists. The service expects a single form part with a csv

test/unit/test_visual_recognition_v4.py
```diff
@@ -0,0 +1,851 @@
...
+        service = ibm_watson.VisualRecognitionV4('YYYY-MM-DD',
+                                                 authenticator=authenticator)
+        service.set_service_url(base_url)
+
+        with open(
+                os.path.join(os.path.dirname(__file__),
+                             '../../resources/cars.zip'), 'rb') as cars:
+            detailed_response = service.analyze(
+                collection_ids='collection_id1, collection_id2',
+                features=AnalyzeEnums.Features.OBJECTS.value,
```
mediumTaj: features and collection_ids should be list of strings, not csv. We need to convert these into csv in the service method

PR #696, opened 2019-10-03
Python sdk v4.0.0 release

ehdsouza:
This PR supports release of python SDK v4.0.0


ibm_watson/speech_to_text_v1_adapter.py
```diff
@@ -47,7 +43,6 @@ def recognize_using_websocket(self,
                                   speaker_labels=None,
                                   http_proxy_host=None,
                                   http_proxy_port=None,
-                                  customization_id=None,
```
mediumTaj: why are we getting rid of this? i think its still in the service.

mediumTaj: I asked about this a while ago - even though its deprecated it is still supported in the service. We should keep it in

https://ibm-analytics.slack.com/archives/CKFME903C/p1566841743005000

mediumTaj:
Also, erika can you regenerate Discovery using `ef9010ab7de1b1fc7628a340cab118e2f00dd51d`? There were some issues with the cpd only functionality that was added.

PR #703, opened 2019-11-06
Fix setting url in microphone-speech-to-text example

SiddhantNaik1731:


examples/microphone-speech-to-text.py
```diff
@@ -37,6 +37,7 @@
 # initialize speech to text service
 authenticator = IAMAuthenticator('your_api_key')
 speech_to_text = SpeechToTextV1(authenticator=authenticator)
+speech_to_text.set_service_url('https://gateway.watsonplatform.net/speech-to-text/api')
```
ehdsouza: This url seems incorrect. It should be `https://stream.watsonplatform.net/speech-to-text/api`

PR #714, opened 2019-11-22
v4.1.0

ehdsouza:
related to: https://github.ibm.com/Watson/developer-experience/issues/7602

PR #716, opened 2019-11-24
Travis CI: Python cache is pip, not pip3

cclauss:


ehdsouza:
@cclauss thanks for the PR. Ill wait for https://github.com/watson-developer-cloud/python-sdk/pull/714 before we can merge in this one

.travis.yml
```diff
@@ -42,19 +40,19 @@ deploy:
   script: docs/publish.sh
   skip_cleanup: true
   on:
-    python: '3.5'
+    python: 3.8
```
germanattanasio: I think the idea is to test in the lower version not the latest.

.travis.yml
```diff
@@ -27,10 +26,9 @@ script:
 - tox
 
 before_deploy:
+- pip3 install bumpversion pypandoc
```
germanattanasio: we install packages using `pip3` but the cache expects `pip` 🤔  why>?

cclauss: That is just the want that Travis works.  In a Travis CI job that does pip3 installs, do:
* ls -R $HOME/.cache/pip  # -> a bunch of nested directories
* ls -R $HOME/.cache/pip3  # -> ls: cannot access '/home/travis/.cache/pip3': No such file or directory

.travis.yml
cclauss: This task is not a test.  Here we are publishing docs.  If we can successfully publish our docs under Python 3.8 that this task should continue to work thru the [EOL date of v3.8](https://devguide.python.org/#branchstatus) which is 2024 instead of 9/2020 for v3.5.

PR #725, opened 2020-01-28
Add documentation for getting transaction id

ehdsouza:
related to [#8010](https://github.ibm.com/Watson/developer-experience/issues/8010)


README.md
```diff
@@ -362,6 +363,30 @@ This would give an output of `DetailedResponse` having the structure:
 ```
 You can use the `get_result()`, `get_headers()` and get_status_code() to return the result, headers and status code respectively.
 
+## Getting the transaction ID
+Every response from the SDK will contain a transaction ID. This transaction ID is useful for troubleshooting and accessing relevant logs from your service instance.
```
mediumTaj: Can you update to 
>Every SDK call returns a response with a transaction ID in the x-global-transaction-id header. This transaction ID is useful for troubleshooting and accessing relevant logs from your service instance.

examples/personality_insights_v3.py
```diff
@@ -17,7 +18,7 @@
 
 # Authentication via external config like VCAP_SERVICES
 service = PersonalityInsightsV3(version='2017-10-13')
-service.set_service_url('https://gateway.watsonplatform.net/personality-insights/api')
+service.set_service_url('https://api.us-east.personality-insights.watson.cloud.ibm.com/instances/4c18b521-3abd-4c7c-bec7-6a3fd03644f1')
```
mediumTaj: I don't think you need to show the instance url here. You can use the regular us-east url
```
https://gateway-wdc.watsonplatform.net/personality-insights/api
```

ehdsouza: There is a bug in recent python core: https://github.com/IBM/python-sdk-core/issues/54. Matthew Keezer said he will work on it

PR #744, opened 2020-05-08
Add handling for noisy or silent utterances

lucascerfig:
The condition for non-empty utterance needs to be added since empty or noisy utterances makes Watson return nothing, leading to errors like this:

File "/path/to/lib/websocket/_app.py", line 320, in _callback
    callback(self, *args)
 File "/path/to/lib/ibm-watson/awebsocket/recognize_listener.py", line 199, in on_data
    hypothesis = json_object['results'][0]['alternatives'][0]


ibm_watson/websocket/recognize_listener.py
```diff
@@ -196,16 +196,20 @@ def on_data(self, ws, message, message_type, fin):
...
             hypothesis = ''
             if 'results' in json_object:
-                hypothesis = json_object['results'][0]['alternatives'][0][
-                    'transcript']
-                b_final = (json_object['results'][0]['final'] is True)
-                transcripts = self.extract_transcripts(
-                    json_object['results'][0]['alternatives'])
+                if len(json_object['results']) != 0:
+                    # Handling for noisy/silent utterances
+                    print(json_object['results'])
```
mamoonraja, line 201: do we need the `print` statement here?

ibm_watson/websocket/recognize_listener.py
```diff
@@ -196,16 +196,20 @@ def on_data(self, ws, message, message_type, fin):
         elif 'results' in json_object or 'speaker_labels' in json_object:
             hypothesis = ''
             if 'results' in json_object:
-                hypothesis = json_object['results'][0]['alternatives'][0][
-                    'transcript']
-                b_final = (json_object['results'][0]['final'] is True)
-                transcripts = self.extract_transcripts(
-                    json_object['results'][0]['alternatives'])
+                if len(json_object['results']) != 0:
```
mamoonraja, line 199: Wondering if we can do something `if 'results' in json_object and if len(json_object['results']) != 0` instead.

ibm_watson/websocket/recognize_listener.py
lucascerfig: The print statement is not needed, it can be removed or replaced.

ibm_watson/websocket/recognize_listener.py
lucascerfig: It does really seems a cleaner and more robust approach. I'd like to emphasize that the changes I proposed were the same changes we needed in our project to make it work, any propositions and edits from this repo maintainers will be happily accepted.

germanattanasio:
@lucascerfig do you think you can update the PR with the requests from @mamoonraja? I think it will be great to get this PR merged.

mediumTaj:
@mamoonraja is this PR ok to merge?

ibm_watson/websocket/recognize_listener.py
mamoonraja: Sorry forgot about this PR, if we get rid of the print statement we should be good to go.

PR #747, opened 2020-05-28
Feat/release 4 2020

mamoonraja:
Generate SDK based on:

```
API def: sdk-2020-05-21
Generator: 2.2.5
```


ibm_watson/assistant_v2.py
```diff
@@ -1215,17 +1363,19 @@ class LocaleEnum(Enum):
...
-    :attr dict system: (optional) System context data used by the skill.
+    :attr MessageContextSkillSystem system: (optional) System context data used by
+          the skill.
     """
 
     def __init__(self,
                  *,
                  user_defined: dict = None,
-                 system: dict = None) -> None:
+                 system: 'MessageContextSkillSystem' = None) -> None:
```
mediumTaj, line 1378: Was this the change we were avoiding?

test/integration/test_visual_recognition_v4.py
```diff
@@ -116,7 +117,7 @@ def test_04_objects_and_training(self):
         # assert add_images_result is not None
         # image_id = add_images_result.get('images')[0].get('image_id')
         # assert image_id is not None
-        image_id = '1280px-Giraffe_Ithala_KZN_South_202349062c2307571a3f7edc71fe819f'
+        image_id = 'giraffe_00_202349062c2307571a3f7edc71fe819f'
```
mediumTaj, line 120: Any way to add a new image to the collection and add training data to that image and deleting that image instead of changing the training on an image in our current collection?

ibm_watson/assistant_v2.py
mamoonraja: ah I thought I removed it everywhere, good catch.

test/integration/test_visual_recognition_v4.py
mamoonraja: let me update

PR #748, opened 2020-06-04
test: refactor vis rec v4 test after new instance

mamoonraja:


test/integration/test_visual_recognition_v4.py
```diff
@@ -128,40 +127,38 @@ def test_04_objects_and_training(self):
...
         assert train_result is not None
         assert train_result.get('training_status') is not None
 
         # training usage
-        # training_usage = self.visual_recognition.get_training_usage(
-        #     start_time='2019-11-01', end_time='2019-11-27').get_result()
-        # assert training_usage is not None
+        training_usage = self.visual_recognition.get_training_usage(
+            start_time='2019-11-01', end_time='2019-11-27').get_result()
+        assert training_usage is not None
```
mediumTaj, line 157: Does this pass? its looking only in nov 2019

mamoonraja: haha, it's still passing somehow. Good catch. Probably cause we are only asserting `assert training_usage is not None`.

PR #774, opened 2021-01-08
feat: upate core to use 3.3.6

mamoonraja:


mediumTaj (approved):
👍 looks good as long as there are no breaking changes!

PR #782, opened 2021-03-26
fix: return_ did not accept List[str] and was inconsistent with the d…

dhruvbhogle:
fix: return_ did not accept List[str] and was inconsistent with the docstring.

The `return_` arg in some of the query defs should accept  a `List[str]` but currently only accepts `str`. The docstrings however specify that the user can provide a comma separated list which is inconsistent and raises errors like below when a return_ arg is specified

```
ERROR:root:Invalid query syntax '(' at index 4 in 'List('
Traceback (most recent call last):
  File "/root/anaconda3/lib/python3.7/site-packages/ibm_cloud_sdk_core/base_service.py", line 157, in send
    response.status_code, error_message, http_response=response)
ibm_cloud_sdk_core.api_exception.ApiException: Error: Invalid query syntax '(' at index 4 in 'List(', Code: 400 , X-global-transaction-id: 606d7d58-4054-4a33-8b9f-3565b9ac3668
Encountered exception: Error: Invalid query syntax '(' at index 4 in 'List(', Code: 400 , X-global-transaction-id: 606d7d58-4054-4a33-8b9f-3565b9ac3668
```

This fix corrects that and allows the user to specify which part of the document hierarchy they require (e.g:  `["result_metadata", "extracted_metadata"]` and the discovery response payload only returns these fields.

Changes were based off the `def query_notices()` function which seems to accept a List[str] argument successfully.

### DIFF
--- a/ibm_watson/discovery_v1.py
+++ b/ibm_watson/discovery_v1.py
@@ -1722,7 +1722,7 @@ class DiscoveryV1(BaseService):
               passages: bool = None,
               aggregation: str = None,
               count: int = None,
-              return_: str = None,
+              return_: List[str] = None,
               offset: int = None,
               sort: str = None,
               highlight: bool = None,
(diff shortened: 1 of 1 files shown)


mediumTaj:
@apaparazzi0329 can you take a look

apaparazzi0329:
Sure thing

apaparazzi0329:
It appears this is something we will want to bring up with the api definitions team as to why `return_` is used as a string and not a list. All of our sdks are auto generated using our Watson API definitions so any changes to this property will need to be done there, internally. Thank you, @dhruvbhogle for bringing this to our attention. Closing PR as this issue will be resolved internally.

PR #783, opened 2021-03-30
fix(discov2): fix query_notices method to include correct path params

apaparazzi0329:


repjarms:
Waiting for some discussion on official API definition updates.

PR #789, opened 2021-05-20
Release 2021 05 19

apaparazzi0329:


repjarms (commented):
It's looking good - but I have a couple questions before approval.

ibm_watson/assistant_v1.py
```diff
@@ -9831,17 +10422,19 @@ class DialogNodeOutputGenericDialogNodeOutputResponseTypeConnectToAgent(
...
-        transfer_info: 'DialogNodeOutputConnectToAgentTransferInfo' = None
-    ) -> None:
+            self,
+            response_type: str,
+            *,
+            message_to_human_agent: str = None,
+            agent_available: 'AgentAvailabilityMessage' = None,
+            agent_unavailable: 'AgentAvailabilityMessage' = None,
+            transfer_info: 'DialogNodeOutputConnectToAgentTransferInfo' = None,
+            channels: List['ResponseGenericChannel'] = None) -> None:
```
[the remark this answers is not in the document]
apaparazzi0329: I do not know but it seems to be standard for init params to have this much indentation. Also, the extra indentation for channels does not show up in my local code

ibm_watson/natural_language_understanding_v1.py
```diff
@@ -1937,29 +3486,37 @@ class Features():
...
+    :attr SummarizationOptions summarization: (optional) (Experimental) Returns a
+          summary of content.
+          Supported languages: English only.
     :attr SyntaxOptions syntax: (optional) Returns tokens and sentences from the
           input text.
     """
 
     def __init__(self,
                  *,
+                 categories: 'CategoriesOptions' = None,
```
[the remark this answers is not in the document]
apaparazzi0329: Ahh thanks for the reminder I believe that I forgot to regenerate NLU after the positional parameter change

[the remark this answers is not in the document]
apaparazzi0329: Ok this has been address in the latest commit

ibm_watson/natural_language_understanding_v1.py
```diff
@@ -1947,19 +3495,24 @@ class Features():
 
     def __init__(self,
                  *,
+                 classifications: 'ClassificationsOptions' = None,
                  concepts: 'ConceptsOptions' = None,
                  emotion: 'EmotionOptions' = None,
                  entities: 'EntitiesOptions' = None,
                  keywords: 'KeywordsOptions' = None,
-                 metadata: object = None,
+                 metadata: 'MetadataOptions' = None,
```
repjarms, line 3503: are we allowed to change this from an object to `MetadataOptions`

ibm_watson/natural_language_understanding_v1.py
```diff
@@ -1947,19 +3495,24 @@ class Features():
 
     def __init__(self,
                  *,
+                 classifications: 'ClassificationsOptions' = None,
```
repjarms, line 3498: If these parameters are fully positional and not key based then I'm not sure we can put this at the top of the list. Can you confirm if these are used positionally or by key?

ibm_watson/natural_language_understanding_v1.py
apaparazzi0329: MetadataOptions itself is just an empty object. No code changes should result from this change although it is still an odd one. This is likely to implement new properties in the future

ibm_watson/natural_language_understanding_v1.py
apaparazzi0329: So python offers both positional and key based parameters. If no keys are given, Python will input them in order so this change would force people to use key based parameters for the Features object

apaparazzi0329: Most of the SDKs already have this order implemented. Should I reverse it for python?

repjarms (approved):
LGTM

PR #794, opened 2021-06-24
Switch deploy methods

apaparazzi0329:
This branch currently only uploads to test.pypi.org


Mikemosca (approved):
LGTM, in deploy don't forget to switch back to workflow_run.  For the pypi action you are still going to test.pypi

PR #795, opened 2021-07-19
ci(release): change release commit messages to include skip c_i

apaparazzi0329:
This PR should stop double GHA release builds from triggering by appending [skip ci] to their commit messsages


Mikemosca (approved):
Very good!

PR #799, opened 2021-08-18
fix(nlc): add deprecation warning

Hsaylor:
add deprecation warning to NLC constructor

fix #9624


ibm_watson/natural_language_classifier_v1.py
```diff
@@ -84,6 +84,15 @@ def classify(self, classifier_id: str, text: str,
         :rtype: DetailedResponse with `dict` result representing a `Classification` object
         """
 
+        print(
```
apaparazzi0329, line 87: This print statement should be within the initializer block above on line 48

apaparazzi0329 (changes requested):

PR #804, opened 2021-09-15
chore: update CHANGELOG

apaparazzi0329:


kevinkowa (approved):
Looks good, but should we emphasize the changes more? I believe the java sdk looks like this, but we want to add another line explaining the "breaking changes"

apaparazzi0329:
In that sense, I think every bug fix would be considered a "breaking change" since users will have to change their code to address the fix. Since these are listed as bug fixes though I don't think we need to specify that they are breaking changes as well

apaparazzi0329:
A user will still be able to come to the changelog and see the comment that relates to their error and be able to look at the commit change to figure out what they need to do

kevinkowa:
Alright, that's fine 👍

PR #822, opened 2022-08-10
build(version): prep version numbers for semantic release fix

apaparazzi0329:


Hsaylor (approved):
doing a heck of a job bumping that version

PR #837, opened 2023-10-18
ibm docs link

krishvsoni:
I had added url of IBM cloud account and Watson API for users who are new they can get direct link for that and it would be easy for setup.


krishvsoni:
Is it considerable ?

